PR #200, opened 2010-10-01
Stable

dparis:
Declaring UTF-8 in _font-face.scss allows SASS to successfully compile the file on systems where UTF-8 is not the default. Been running this without incident for a week or two.


chriseppstein:
merged. thanks.

PR #212, opened 2010-10-29
stretch, stretch-x, and stretch-y mixins for absolute positioning

steve-wong:
Some simple mixins to stretch an element via absolute positioning.


steve-wong:
Ahh...sorry for jumping the gun.  Thanks for the link to the instructions on contributing.  I'll work on getting tests, docs, etc. and changing the implementation so 'position: absolute' isn't output twice.

I can also change the name to "stretcher" as suggested.  Yes, it can be used to create backdrops, but is also useful for absolutely positioning a panel or viewport area.  For instance, stretching a div to fill the right side (or some portion) of a container.  It can be an alternative to percentage based width/height.

chriseppstein:
Merged to master: https://github.com/chriseppstein/compass/compare/a618ff3509c446b4a1a1...a8a2f42b

chriseppstein:
thanks!

PR #214, opened 2010-10-29
IE8 compatibility with :last-child selectors

jdsiegel:
This request separates any rules within compass that include a :last-child selector. IE8 with IE8 standards will ignore all selectors that are on the same line as a :last-child selector. Separating :last-child from these selectors fixes the issue.

More information can be found here: http://www.richardscarrott.co.uk/posts/view/ie8-last-child-bug


chriseppstein:
I cherry picked this to stable. Thanks!

http://github.com/chriseppstein/compass/commit/762c6a0b188efda2d4a8575fed64988d97bc6045

Issue #208, opened 2010-10-24
Add Opera and IE9 support for gradients via SVG


alanhogan:
Given
- IE9 (as of the last beta I checked) supports _neither_ CSS3 gradients _nor_ `-ms-filter` gradients;
- Opera still lacks CSS3 gradients;
- Both Opera and IE9 allow SVGs to be used as backgrounds (and also support `data:` URIs); and
- Compass already includes a feature to “inline” and base64-encode images,

I think it would make sense for Compass to generate SVG gradients, base64 compress them, and set them as the background, for these browsers.

The SVG can be pretty simple. Here is a vertical gradient with one semi-opaque stop.

```
<?xml version="1.0" ?> 
<svg xmlns="http://www.w3.org/2000/svg"> 
 <defs>
  <linearGradient y2="1" x2="0.5" y1="0" x1="0.5" id="svg_gradient_2">
   <stop stop-color="#000000" offset="0"/>
   <stop stop-opacity="0.5" stop-color="#ffffff" offset="1"/>
  </linearGradient>
 </defs>
 <g>
      <rect id="svg_1" x="0" y="0" width="100%" height="100%" fill="url(#svg_gradient_2)"/>
 </g>
</svg>
```

alanhogan:
Oh, and the SVG can be further compressed:

```
<svg xmlns="http://www.w3.org/2000/svg"> 
  <linearGradient y2="1" x2="0.5" y1="0" x1="0.5" id="svg_gradient_2">
   <stop stop-color="#000000" offset="0"/>
   <stop stop-opacity="0.5" stop-color="#ffffff" offset="1"/>
  </linearGradient>
  <rect id="svg_1" x="0" y="0" width="100%" height="100%" fill="url(#svg_gradient_2)"/>
</svg>
```

alanhogan:
Added benefit: Gradients that actually rely on published non-draft web standards!

alanhogan:
Related: This tool generates SVG and PNG fallbacks along with the traditional options. 

http://www.display-inline.fr/projects/css-gradient/#startType=rgba&extOpera=1&extIE=1

chriseppstein:
I think this makes sense as an opt-in feature. Anyone want to try their hand at a patch?

chriseppstein:
related: #183

dira:
I wrote opt-in support for linear gradients, but I got stuck when adding a new stylesheet for testing it, as it makes integration tests fail.

[...]

PR #215, opened 2010-11-06
Gradients: with SVG implementation (issue #208)

dira:
I added support for gradients with SVG backgrounds (issue #208). It uses the same functions as for Webkit to calculate the stop points.

I tested it thoroughly in Opera, and not at all in IE9 (I don't have it :) ).

### DIFF
--- a/frameworks/compass/stylesheets/compass/_support.scss
+++ b/frameworks/compass/stylesheets/compass/_support.scss
@@ -23,3 +23,5 @@ $experimental-support-for-opera     : true !default;
 $experimental-support-for-microsoft : true !default;
 // Support for khtml in experimental css3 properties.
 $experimental-support-for-khtml     : true !default;
+// Support for svg in experimental css3 properties.
+$experimental-support-for-svg     : false !default;
--- a/frameworks/compass/stylesheets/compass/css3/_gradient.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_gradient.scss
@@ -1,30 +1,30 @@
 @import "shared";
 
-// This yields a linear gradient spanning from top to bottom
+// The linear gradient mixin works best across browsers if you use percentage-based color stops.
 //
+// Examples:
+//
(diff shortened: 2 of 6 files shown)


dira:
I "tested" it on IE9 beta as well, through BrowserCam, and the test page renders correctly: [http://www.browsercam.com/public.aspx?proj_id=533356](http://www.browsercam.com/public.aspx?proj_id=533356)

The test page is [http://dira.ro/code/CSS3-gradients/index.html](http://dira.ro/code/CSS3-gradients/index.html), which was generated with the svg-enabled linear-gradient and radial-gradient mixins, and works in Webkit, Firefox and Opera.

chriseppstein:
It's merged to the compass css3-v2 branch which will land on master soon as part of a css3 module revamp. Thanks for the patch, it's great. Also: did you know that svg backgrounds don't seem to render on elements with a border in opera? That tripped me up for a while during testing :(

dira:
**About the percentages:**

SVG works with percentages only, which makes it very similar with the Webkit implementation for linear gradients. 

For radial ones, I set the radius of the circle to fill; then, use again the code for Webkit to get percentages for the color stops. So it's not by choice, but rather: the only implementation that is feasible behaves identical to Webkit. :D

**About Opera:**

Yes, I had trouble with that bug as well, and sent Opera a report about that. But, I have no idea what's the status, as their bug system is very closed.

For what I encountered, elements with border show if they are in the first screen, and if you scroll them out of view they become empty.
If you put a border AND `background-size: 100% 100%`, it will kind-a-work, rendering bad at the end of the gradient.
See the demo page [http://dira.ro/code/Opera-bug/](http://dira.ro/code/Opera-bug/)

Looking again at it, I think compass could just set background-size to 100%. This should be neutral to other browsers and prevent this bug on Opera.

I will test this later this weekend, to confirm that it does not cause other problems, and commit the patch if it's ok.

chriseppstein:
Sounds great. Thanks for the follow up.

PR #218, opened 2010-11-13
Fix issue 199: added font configuration options 

rockdog:
Added font configuration options to configuration reference. Default values for http_fonts_(dir|path) are missing.


chriseppstein:
rebased and merged to master. Thanks!

PR #219, opened 2010-11-15
Fix SVG gradients on Opera

dira:
I checked setting `background-size: 100%` and it works great on all browsers (only Opera takes it into account for the moment).

In the note about Opera I recommend setting `no-repeat` on the background. This generally makes the rendering problem more "acceptable" on Opera, as it just leaves an unfilled margin instead of repeating the gradient: http://dira.ro/code/CSS3-gradients/ .

### DIFF
--- a/doc-src/content/reference/compass/css3/gradient.haml
+++ b/doc-src/content/reference/compass/css3/gradient.haml
@@ -13,7 +13,11 @@ classnames:
 - render 'reference' do
   %p
     Provides mixins to create cross-browser CSS3 gradients.
+  %p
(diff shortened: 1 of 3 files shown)


chriseppstein:
cherry-picked to master. Thanks!

timhettler:
This change can cause issues when you pass multiple background images to the mixin. (Specifically, a gradient + a repeating texture.) Can we re-evaluate whether this Opera bug still needs to be accommodated for?

Example here: http://codepen.io/timhettler/pen/uEpzj

PR #232, opened 2010-12-04
Extra sass functions

rdougan:
Thought the first-value-of method was handy, so I added 1st, 2nd, 3rd and 4th.


chriseppstein:
sass 3.1 provides a generic function for this: `nth($list, $index)`

The helper function was just until sass 3.1 is out, so I won't be pulling in this changeset. Thanks for the patch though.

PR #235, opened 2010-12-06
sass dependecy fix

slbug:
after today's sass version bump i got this problem

<pre>
Bundler could not find compatible versions for gem "sass":
  In Gemfile:
    compass depends on
      sass (>= 3.1.0.alpha.50)

    sass (3.1.0.alpha.0)
</pre>


chriseppstein:
Thanks for the patch, but the compass version dependency is correct. The problem is that the sass master branch does not correctly identify itself.

PR #245, opened 2011-01-12
Added a reset-float mixin

adamstac:
I've used this often enough and thought, why the heck don't I add this to Compass.

There have been a number of cases where I wanted to reset a float created by one of the float mixins such as `+float-left`. Before I wrote this, I would have to set the element to `float: none` and then also set `display: block` to reset the changed display. Using this mixin, you would use `+reset-float` to reset the float and if the element is inline you can pass the argument 'inline' to keep the element's display set to inline, for example `+reset-float(inline)`.


adamstac:
Sent a new pull request. This one is invalid now.

AS

PR #246, opened 2011-01-12
float-reset updated

adamstac:
Updated the code to what hagenburger and I commented about in my fork. See this pull request for more details: https://github.com/chriseppstein/compass/issues#issue/245
- Adam


scottdavis:
probably want this to be against master and not stable 250 commits is alot =) I made this mistake once

chriseppstein:
I've merged this change into master. I moved the mixin in the float module because you only need it when you're using the other mixins in that module.

chriseppstein:
I've merged this change into master. I moved the mixin in the float module because you only need it when you're using the other mixins in that module.

chriseppstein:
https://github.com/chriseppstein/compass/commit/9b0967eb474c791ff5be7b0510cfdb5a01db6c46

PR #257, opened 2011-01-20
Opps

scottdavis:
updated tests and bumped the chunky_png version to 0.12.0 also updated git ignore to ignore Gemfile.lock since i goofed that all up hence the branch name


chriseppstein:
took a diff one.

PR #269, opened 2011-02-03
here is the generalized delimited-list

chadoh:
I don't know the process for marking stylesheet features deprecated, so I just noted it in the comments above "comma-delimited-list" (it now does the same thing as "delimited-list" with no argument, so it seems fairly useless to me).

One of the tests causes compass to throw a deprecation warning. I didn't add any scss files or touch the tests so I assume it wasn't me.

The cucumber tests gave all sorts of failures. I assume I have no responsibility for that. :-\


chriseppstein:
merged to master. Thanks!

PR #275, opened 2011-02-17
added callbacks support using Sass::Callbacks

scottdavis:


scottdavis:
checkout my current master it has the failing test you said you would look into

chriseppstein:
merged to master

PR #278, opened 2011-02-20
Updated the Callbacks to use the method_missing so that the linked list structure doesn't get in the way

scottdavis:
All tests run and pass


chriseppstein:
this is merged to master now.

PR #280, opened 2011-02-20
Call backs not moving down from top of the linked list

scottdavis:
In the way you implemented if you add config options before the project config is loaded those callbacks never get called. Because call backs only move down the list and not up events should start from the top and move the entire way down the list since it is possible to make call backs defined on "n" config objects. This seems like a crud way todo it in my patch here but it gets the point across what may need to be adjusted.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    compass (0.11.beta.2.0dc38bc)
+    compass (0.11.beta.2.8a8ef87)
       chunky_png (~> 0.12.0)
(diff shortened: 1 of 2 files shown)


scottdavis:
https://gist.github.com/81cd9d7c50997767e10c

if i execute this before a project config loads and save a file that proc never gets ran

but if i define the callback proc in the project config it gets ran

chriseppstein:
closing... not necessary afterall.

scottdavis:
Im still having a hard time settling with the way we talked about in IRC maybe we need global callbacks also its not as seamless as i had hoped integration would be

PR #283, opened 2011-02-22
Webkit linear-gradient support

stevelacey:
Webkit are now in line with the spec for linear-gradient support, so I have updated this mixin to accommodate.

I've changed the properties to 'background', not sure why but my styles at least wouldn't work on 'background-image'. I imagine it's either to do with multiple background or gradient support? Either way, changing it fixed it and I have no desire to continue my examination as to why it did.

Webkit have also brought radial-gradient in line with the spec, this should also be updated at some point. Again, I don't need this right now so haven't made the changes myself, they are invariably similar however except that the start/end positions are set as the first parameters and then the colours last rather than position, colour pairs, perhaps it would be wise to simply accept one quoted parameter to allow for full flexibility, I think I'm right in saying most modern browsers that support gradients are now using the same syntax anyway, thus we only need vendor replication.

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_gradient.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_gradient.scss
@@ -2,29 +2,29 @@
 
 // This yields a linear gradient spanning from top to bottom
 //
-//     +linear-gradient(color-stops(white, black))
+//     +linear-gradient("white, black")
 //
 // This yields a linear gradient spanning from bottom to top
 //
-//     +linear-gradient(color-stops(white, black), bottom)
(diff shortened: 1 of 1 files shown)


chriseppstein:
Thanks for the patch, but on master this is completely re-written. Also the old syntax continues to be supported and should remain for legacy versions of webkit for at least a while. When the browsers drop their prefix for this property they will pick up their correct version that is already there. So I don't think any changes are necessary at this time.

stevelacey:
Yes, I realised my failure minutes after submission.

Ta for the response,

Steve

totty90:
+1

PR #293, opened 2011-03-02
Sprite abstraction

scottdavis:
feel free ot be as harsh as possible im still unsure of a few things like file placement etc all tests are green and we abstracter out the hash of image settings into its own class so its more sain to read the code =). i also refactored the entire base class.

i don't want you to merge this one just to review / comment on implementation since there is still some grossness like

https://github.com/jetviper21/compass/blob/master/lib/compass/sass_extensions/functions/sprites.rb#L16

lingering around


scottdavis:
DAMNIT

wvanbergen:
Just a small heads up: with the imminent ChunkyPNG 1.0 release, you probably want to use #replace! instead of #replace, because #replace will return a new canvas instead of changing the current one. I also implemented a C version of the #replace! method in OilyPNG, so it should be a lot faster as well. 

You can test the new version by installing ChunkPNG/OilyPNG 1.0.0.rc2

PR #301, opened 2011-03-16
transform-style mixin corrections

doctyper:


mirisuzanne:
Made the change here, along with the needed test update, and pushing to master rather than stable. Thanks @doctyper!

Fixed by commit: ddea94bb9650db0aaeaa

doctyper:
Knew I screwed that request up somehow.
Oh, I think you missed the second part of that commit. The `preserves-3d` value should be `preserve-3d`.

mirisuzanne:
good catch. screw-up now mutual, and fixed. :-)

PR #312, opened 2011-03-28
Enhanced cache busting by allowing path manipulation

nbibler:
I modified the cache busting routine to allow for path modification.  This is incredibly useful for CloudFront, for example, which entirely ignores query string parameters - so the default mtime approach has no use.  Traditionally, the approach here is to modify the path by moving some dynamic value into the request path, like a timestamp, mtime, MD5 hash, deploy identifier, or by some other means.

The modification adds another layer between the URL generator and the cache busting generator.  It calls the cache buster and checks for a `/` (forward slash) in the response.  If found, it is assumed that you're doing a path manipulation, rather than a simple query string append.  So, the old append still works, but if a `/` is returned, it will entirely replace the previous path.

Tests were added, documentation was updated.  The "patch" was made against the current, `stable` branch.

Thanks,
Nate


scottdavis:
Have up tried it ago at the current master? Alot has changed since we are getting close to 0.11

Scott Davis

nbibler:
Honestly, I didn't even notice that I was on stable (I guess that's the GitHub default branch) until I was cleaning up for the pull request.  I'll try to apply the functionality to master, as well.

nbibler:
Closing this pull request, a second was made against the `master` branch at https://github.com/chriseppstein/compass/pull/313

PR #313, opened 2011-03-28
Enhanced cache busting by allowing path manipulation

nbibler:
I modified the cache busting routine to allow for path modification.  This is incredibly useful for CloudFront, for example, which entirely ignores query string parameters - so the default mtime approach has no use.  Traditionally, the approach here is to modify the path by moving some dynamic value into the request path, like a timestamp, mtime, MD5 hash, deploy identifier, or by some other means.

The modification adds another layer between the URL generator and the cache busting generator.  It calls the cache buster and checks for a `/` (forward slash) in the response.  If found, it is assumed that you're doing a path manipulation, rather than a simple query string append.  So, the old append still works, but if a `/` is returned, it will entirely replace the previous path.

Tests were added, documentation was updated.  The "patch" was re-made against the current, `master` branch.

Thanks,
Nate


chriseppstein:
It seems like this approach for deciding the mode results in an inconsistent behavior for the degenerate case:

   background: image-url("logo.png");

Where the intent would be to serve something like:

```
background: url("/images/logo-a2e75a.png");
```

I think the way to handle it is to make the mode explicitly part of the return value for the block. For instance they should return:

```
asset_cache_buster do |path, file|
  {:path => "#{...}", :query => nil}
end
```

And then compass would interpret a string response as meaning:

```
{:path => nil, :query => return_value }
```

Thoughts?

nbibler:
Sure.  I would agree that I'm not thrilled with the pattern matching to decide functionality.  However, there were a few thoughts going into this:
1. I didn't want to break backward compatibility, especially since this isn't scheduled to be a major release, and
2. To address @chriseppstein on inconsistency, I don't believe that it's inconsistent, since people who are in a situation where they are explicitly overriding the cache buster to modify the path will recognize and understand the modified path and understand that it does not directly equal the requested `image-url`, for example.  Again, the "traditional" method for this type of busting is to make a fake subpath in the request that is caught by either the web server or a middleware and removed prior to being served, as an example, not necessarily modify the requested filename, [GitHub is currently doing this](https://github.com/blog/815-hello-singapore-cdn-experiment) by inserting a Git SHA as the base request path and having nginx interpret it.  The example given and test written was just a proof of concept that the path - not the query string - was modifiable.

I agree that it would be better to be more explicit going forward.  But, I can't think of a case where pattern matching will fail.  Since, you've already stipulated that the cache buster block must return either a string or nil which will be appended to the query string, people **should** already be escaping non-query string compatible characters (slash, colon, space, etc.).  So, I think it's fair to assume that if it contains an unescaped forward slash, then they want to modify more than the query string.

I'm certainly open to modification... I don't really appreciate the pattern match, but it seemed to be the shortest distance without breaking compatibility or being overly intrusive on the current path-generation process.  Ideally, I'd like to see Compass be smart enough to recognize that it's in a Rails environment and just use the Rails `asset_host` and `asset_path` string/block and then this is all a fairly moot point.  That would reduce my code duplication (between the environment configuration and the Compass initializer).

Since I need this for a current project, I'm currently using my fork with this modification in production.

chriseppstein:
@nbibler I'm pretty sure your code doesn't work for the simple case I noted earlier. the patten matching approach is ambiguous in the case where the image is in the top level of your images directory. Because of that, we need the hash-based return value if they want to return a path change. Happy to merge this pull request after that change is done.

donaldpiret:
This functionality would be very very very nice to have, whatever way it's implemented

pfeiffer:
Totally agree! :-)

chriseppstein:
If only +1's were patches :) It's on the list of things to do, but if it's blocking you, please scratch your itch.

pfeiffer:
Added a patch to support `{:path => .., :query => ...}` in #372 :-)

chriseppstein:
I'll check it out. Thanks!

chriseppstein:
closing this pull request in favor of #372.

PR #315, opened 2011-03-28
Check if Rails module responds to env and root before using them

zmack:
Hey there,

This helped me get compass up and running on Rails 2.0.5, and it's pretty much forward compatible.

Hope it helps.


chriseppstein:
I made a small bug fix, so this is applied to the master branch as c4a315af3c3b7adf4c4c0d8df3678e6f75e946f5. Thanks.

PR #319, opened 2011-03-29
linear-gradient now support Opera 11.10 beta

WaYdotNET:
I hope this patch is good for all


chriseppstein:
The code on master is very different, I already have a patch in the works for this issue and it'll be fixed on the master branch.

PR #321, opened 2011-03-29
Forcewrapmixin

gmclelland:
Added force-wrap mixin to prevent URLs and long lines of text from breaking layouts.


gmclelland:
Note: This needs to be applied to block level elements.  I got this from http://perishablepress.com/press/2010/06/01/wrapping-content/  so we probably also need to include that in a comment.

gmclelland:
There you go as requested.  Hope that helps

PR #326, opened 2011-04-03
add support for type=tel input fields

bonkydog:
Hi, Chris.

Thanks for Compass!

I was building a form and found the telephone number field was looking a little wonky.

Found that compass wasn't formatting inputs of type "tel" (though it does "email" and "password").

So forked and added support for inputs of type "tel".  I'd like to share this little fix with others.

Please pull it into the canonical compass repo?

Best,
Brian

### DIFF
--- a/frameworks/blueprint/stylesheets/blueprint/_form.scss
+++ b/frameworks/blueprint/stylesheets/blueprint/_form.scss
@@ -22,6 +22,7 @@
     &.text,
     &.title,
     &[type=email],
+    &[type=tel],
     &[type=text],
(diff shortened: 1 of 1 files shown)


chriseppstein:
Hi,

Thanks for the patch, but I only make changes to the compass port of blueprint that are in the blueprint library. Is this there already and I missed it, or does it still need to be added to blueprint?

bonkydog:
Hi, Chris.

No, it's not in blueprint.  I'll talk to them about adding it.

Thanks!
Brian

PR #372, opened 2011-05-04
Cache busting with path and query

pfeiffer:
Related to #313:

This patch is enables asset_cache_buster to return a hash with :path and/or :query values. These values are used to generate the busted asset path.

Backwards-compatability has been preserved, meaning a string value returned by asset_cache_buster will result in a query string parameter.

I've added some additional tests.


test/fixtures/stylesheets/busted_image_urls/sass/screen.sass
```diff
@@ -0,0 +1,14 @@
+.showgrid
+  background-image: image-url(unquote("grid.png"))
```
chriseppstein: These unquotes shouldn't be necessary.

chriseppstein:
This is great. I'm super busy with work, so if you can add a changelog entry and update the configuration docs, I'll merge in for the release this weekend.

pfeiffer:
Added changelog and updated the configuration docs :-)

agibralter:
+1 \* 100!

I'm trying to get compass to work with https://github.com/donaldpiret/asset_hash, which fingerprints assets with md5 hashses with the pattern `file_name-#{hash}.ext` which plays nice with CDNs like CloudFront...

donaldpiret:
agibralter, you can nbibler's fork of compass at the moment which plays nice with asset_hash (it's what we're using here):

gem 'compass', '0.10.6', :require => 'ninesixty', :git => "git://github.com/nbibler/compass.git", :branch => 'busting'

Or if you need compass '0.11' you can use my custom branch which is basically nbibler's changes merged into master

gem 'compass', '0.11.1.32a05b0', :require => 'ninesixty', :git => "git://github.com/donaldpiret/compass.git", :branch => 'busting-master'

agibralter:
@donaldpiret Hmm so I ended up writing a Rake tast that post-processes my css files: https://gist.github.com/991732

I use jammit too, which tries to do it's own asset path rewriting...

Basically, I think all theses libraries should have more flexible configuration options for either turning off or turning on asset path rewriting. Ideally they would all just plug into Rails's `asset_path=` and `asset_host=` procs.

chriseppstein:
Clearly I lied about the release :P But I will merge this in soon.

PR #373, opened 2011-05-05
Dynamically calculate Blueprint outer-width and container so that @import order doesn't change the output 

tdreyno:
I've got this bug report on middleman:
https://github.com/tdreyno/middleman/issues/45

Basically, depending on whether you @import blueprint before or after defining some variables, the output is different. Given the !default flag, you would think that you could update the variables and have the output change.

Following susy's lead, I switched the internal variables calculated on @import into @functions that take the public variables into account. For backwards compatibility, the "private" variables, if set, will still be used.

### DIFF
--- a/frameworks/blueprint/stylesheets/blueprint/_grid.scss
+++ b/frameworks/blueprint/stylesheets/blueprint/_grid.scss
@@ -27,11 +27,30 @@ $blueprint-grid-width: 30px !default;
 // The amount of margin between columns
 $blueprint-grid-margin: 10px !default;
 
+// Directly set the outer width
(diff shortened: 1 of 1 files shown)


chriseppstein:
I like what this commit accomplishes, but I fear it makes something that is overly complicated even more complicated.

tdreyno:
I agree :)

In fact, that's why I use Susy. Maybe I'll just switch middleman to include Susy and change all the demos/templates to use that instead. The current Blueprint implementation is confusing... but it's stuck between two worlds.

chriseppstein:
Closing this without merge. Something needs to be done with blueprint, but I think it's more comprehensive than this.

PR #386, opened 2011-05-12
Adding Micro Clearfix

DavidOliver:
Thought this could be used at some point as it uses less CSS than the PIE clearfix and it seems robust, covering all main browsers. Have put credit in scss. http://nicolasgallagher.com/micro-clearfix-hack/
By the way, this is my first ever fork/pull request. :-)

### DIFF
--- a/frameworks/compass/stylesheets/compass/utilities/general/_clearfix.scss
+++ b/frameworks/compass/stylesheets/compass/utilities/general/_clearfix.scss
@@ -29,3 +29,16 @@
   }
   @include has-layout;
 }
+
+// An update to the PIE clearfix method that reduces the amount of CSS required
+// [A new micro clearfix hack](http://nicolasgallagher.com/micro-clearfix-hack/) (25 April 2011)
+@mixin micro-clearfix {
+  &:before, &:after {
+    content: "";
+    display: table;
+  }
+  &:after {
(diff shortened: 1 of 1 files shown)


chriseppstein:
I'm currently pondering whether we should just cut over pie-clearfix to this and change the current pie-clearfix to legacy-pie-clearfix.

I like to keep compass users up to date without them having to do much work.

Also this needs a test case.

DavidOliver:
Hmm... do you think support for less than v3.5 of FireFox should be
"removed" from the pie clearfix in that way? Wouldn't worry me, but I don't
know what the compass approach to that is.

I'm away until Monday, but does a test case involve any formal procedure?
(Am new to this.)

chriseppstein:
@ericam, @imathis Thoughts?

mirisuzanne:
I like moving to the micro approach in general, and leaving FF < 3.5 to a legacy mixin. 

I don't like the use of the :before element. The only reason for including it is to prevent top-margin collapsing, which is outside the scope of a clearfix, and belongs somewhere else if people want it. I use the :before and :after pseudo-elements enough for other purposes that I don't like them being used up when they don't need to be.

chrisl8888:
Why was this closed? This seems better than the current @clearfix mixin.

PR #405, opened 2011-05-30
Enable IE 10 support for flexible box with the -ms prefix.

dpogue:
IE 10 supports the CSS3 flexible box properties, with the -ms prefix.

http://ie.microsoft.com/testdrive/HTML5/Flexin/Default.html


dpogue:
I've updated the test case and verified that all tests pass with no failures/warnings.

scottdavis:
Thanks a bunch! now lets have @chriseppstein take a look magic things will happen (with or without unicorns)

PR #422, opened 2011-06-07
Fixed css3 example of background property image URL.

tashian:
Just a one-line fix of one of the CSS3 examples.


scottdavis:
can you change this to be against master please?

PR #425, opened 2011-06-11
Wordpress Integration

devonhumes:
I have added application integration to compass for WordPress.  The integration is very simple as all it had to do was set the default css_dir to the root of the project (I just copied and modified the stand_alone application). I have also added a new framework to add basic, compass compatible WordPress themes.  Both of these themes are very basic and are only intended as starting points for projects.  The commands needed to run everything are:
compass init word_press
compass init word_press --using wordpress
compass init word_press --using wordpress/basic
If you have any questions or if you feel I need to add something let me know.

Devon Humes


frameworks/wordpress/templates/basic/_LICENSE.txt
```diff
@@ -0,0 +1,281 @@
+		    GNU GENERAL PUBLIC LICENSE
```
chriseppstein, line 1: This is a non-starter. I cannot distribute GPL code in compass because compass is MIT licensed. I guess it will have to be a plugin so that plugin can be GPL licensed.

chriseppstein, line 1: Another alternative could be that the installer would instead download the GPL files from some authoritative location

chriseppstein:
Going to close this. Let's turn this into a plugin.

devonhumes:
I think that we could still pull the first commit as that doesn't have any GPL code in it.  It is just app integration with WordPress.  I can also remove the GPL code from the framework.  I have already created a plugin/gem based on the framework code called 'compass-wp'.  If you want me to resubmit my code minus the GPL I can do that.

devonhumes:
Could you please help me to understand what the incompatibilities with the license are?  I have never contributed to an MIT licensed project before, but the Free Software Federation says that they are compatible. http://www.gnu.org/licenses/license-list.html#X11License  and http://www.gnu.org/licenses/license-list.html#Expat  I would really like to understand this so I can avoid future errors.

scottdavis:
+1 for plugin

matteosister:
@devonhumes if I'm not wrong....

the so called mit license is compatible with gpl because it's less restrictive.
It allows you to take the code, modify it and then release it with another licence. Even a closed one. But in fact the code is open so that anyone can see or modify or update it as you did.

The (richard stallman's) gpl license force to re-use of gpl licence in all the derived projects. Even if only a part of code is included.

For example: linux kernel is gpl. So also android must be gpl. and also any linux distribution out there must be gpl.

edit: I was wrong. Android is not gpl but apache licence...

chriseppstein:
IANAL, but my understanding is that GPL is a virus. if you have GPL code, you have to be GPL. MIT is basically do whatever. So GPL projects can use MIT code, but not the other way around.

chriseppstein:
@devonhumes Let's go with a plugin for now and get some community feedback before merging. Also, it's hard for me to maintain this because I don't use it and afaik, none of the core team uses wp.

matteosister:
@chriseppstein ianal too. But...just to be clear...a virus spread a disease. GPL spread a good habit of keeping the code open.
I understand why you (and also google with android) do not want to use gpl.
Anyway I think it's a great thing. And I'm very happy that linux is under that licence and not mit one.

p.s.: compass rocks! All my collegues now can't live without it. KEEP UP THE SUPERB WORK!

devonhumes:
Thanks for the explanation!  I think I had the two switched.  I will go ahead and update the plugin.  Is there anyway to integrate an application through a plugin?  That would be best, because it really is just a small change.  All that needs to happen is to have the ability to override the config file so it compiles to the root of the project not to a sub directory.

devonhumes:
ps @matteosister I totally agree!  I'm converting everyone I meet!

chriseppstein:
@matteosister Not all virus spread disease. Scientists are using viruses to make people better too :) But regardless of your definition, the GPL is designed to force people to be open. in a way that limits freedom. I want closed source/paid products to be able to embed compass without worrying about things (as long as they give us credit).

@devonhumes you can just add these same files to your plugin and require them during the plugin initialization. The compass install should work fine like this:

```
compass -r compass-wordpress init word_press --using wordpress
```

yfeldblum:
People who work on commercial software are often happy to use, _and contribute fixes, features, documentation, ideas, and tips of the hat back to_, those open-source projects that are useful to them.

But they often avoid, like the plague, any open-source projects licensed solely under copyleft licenses. And the communities surrounding the useful open-source projects are all the poorer for it.

Open-source non-copyleft projects have spread very far through every corner of the software world, encouraging a large range of people to contribute back to those projects and so enlarge the open-source world. This could not have happened were these projects copyleft.

Were commercial projects using Compass required to disclose their sources, they would abandon Compass; the programmers building these projects would cease battle-testing and improving Compass; and the Compass community as a whole would suffer as a result.

I for one would prefer Compass to be licensed under a permissive open-source license such as the MIT License.

Linux is a very different because it is an operating system kernel. Few applications out there embed Linux. And applications which make heavy use of Linux are not required to disclose their own sources because the GPL permits closed-source or otherwise non-GPL applications to link to system calls in GPL operating systems.

chriseppstein:
@yfeldblum Compass is MIT licensed that's why I had to reject this patch.

yfeldblum:
My bad. I meant "I for one would prefer Compass to _remain completely_ licensed under a permissive open-source license such as the MIT License" rather than "to _be_ licensed." Compass is awesome - and I'd like to keep using it!

matteosister:
@yfeldblum

> Few applications out there embed Linux.

just one or two... :D

PR #435, opened 2011-06-18
Fixed broken link in Helpers documenation

srbiv:
Broken link fixed to what I think is the correct section of the SASS docs.


chriseppstein:
thanks!

PR #436, opened 2011-06-19
Fix typo in SpriteImporter

spohlenz:
This patch fixes the call to raise in Compass::SpriteImporter.path_and_name, so that the following error is not produced:

```
undefined method `Error' for Compass:Module
```


scottdavis:
already fixed in master

sporkd:
Could this patch apply to the rails31 branch as well, or do we wait until rails31 is merged into master?

scottdavis:
i need to merge master into rails31 i should beable todo that once i get home

PR #437, opened 2011-06-19
Rails 3.1 image (and other) assets

sj26:
Guinea-pigging compass on Rails 3.1, noticed image helpers like replace-text didn't work as expected. They find the images on-disk in the right place, but aren't serving the right URLs.

Modfied Rails 3.1 default configuration to set all the HTTP paths to the asset http path. It should actually change depending on the asset pipeline's settings, really, like whether assets are being served from a different host.

Cheers!

### DIFF
--- a/lib/compass/app_integration/rails/configuration_defaults.rb
+++ b/lib/compass/app_integration/rails/configuration_defaults.rb
@@ -6,28 +6,39 @@ module Compass
         # These methods overwrite the old rails defaults
         # when rails 3.1 is detected.
 
+        # Assets dir is configurable, but rails doesn't provide the default
+        def default_assets_dir
+          ::Rails.application.config.assets.assets_dir.presence || File.join("app", "assets")
+        end
+
+        # These two have special case overrides:
+
         def default_sass_dir
-          File.join("app", "assets", "stylesheets")
-        end
-
-        def default_images_dir
-          File.join("app", "assets", "images")
-        end
-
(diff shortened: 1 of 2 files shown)


sj26:
Yup, as I mentioned it should.

I've added another commit (0fb8ab1) to make it respect the asset pipeline settings. I don't know how to make it respect asset hosts and such yet.

spohlenz:
I could be wrong, but won't this strategy break when working with assets supplied by engines, or using multiple asset locations (e.g. lib/assets and vendor/assets)?

sj26:
@spohlenz they should all cascade together...

spohlenz:
Ah of course. I think I got confused with some other code I'd just read.

sj26:
So that wasn't sufficient to work across all environments. Now it uses the sprockets extensions to find out where an asset should be linked so we get stuff like perform_caching? working. This means the url helpers need to call the configuration (or something) so I've added that in, if only temporarily.

This is working for me from dev to prod on Rails 3.1 rc4.

chriseppstein:
Thanks for the pull request, I know it wasn't a trivial patch. But these helpers are now provided by sass-rails so that anyone can use them (not just compass users). As such, the patch that I will need to make for compass is to simply disable the compass url helpers in that environment.

sj26:
No worries, that's probably where it belongs and just happy to see it all working. :-)

PR #443, opened 2011-06-24
Configurable output path for sprites

coderanger:
Add a configurable output path for images (specifically sprite maps) that can be distinct from the input path. This is mostly useful situations like nanoc, where the output is being assembled into a new directory incrementally.

All tests (including new one for new functionality) are passing.

### DIFF
--- a/lib/compass/configuration.rb
+++ b/lib/compass/configuration.rb
@@ -1,10 +1,12 @@
 module Compass
   module Configuration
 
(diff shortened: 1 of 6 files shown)


coderanger:
I wanted to keet the dir/path split so I could just do project_path="contents" output_path="output" in my nanoc configuration, keeping the sub-folders using the same layout.

scottdavis:
im adding an importer load path for sprites to address this and another issue

scottdavis:
implemented in 9c472b3

PR #445, opened 2011-06-28
The existing border-radius code in compass is kind of ridiculous. 

ghost:
- As far as I know, -ms-border-radius, -o-border-radius, and -khtml-border-radius never actually existed in the wild. They either support nothing or "border-radius".
- Webkit has supported shorthand syntax for over a year and a half and hasn't needed a vendor prefix for nearly as long.
- At this point the only browser needing a prefix with a current user base is FF3.6, well on the way out already.

In terms of syntax, browser will take pretty much any value you throw at them, so there's no need to even complicate it by trying to capture parameters. In fact at this point border-radius is one of the blessed few css3 properties that doesn't need a pre-processor to make it manageable.

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_border-radius.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_border-radius.scss
@@ -2,134 +2,32 @@
 
 $default-border-radius: 5px !default;
 
-// Round all corners by a specific amount, defaults to value of `$default-border-radius`.
-//
-// When two values are passed, the first is the horizontal radius
-// and the second is the vertical radius.
-//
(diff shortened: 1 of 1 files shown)


scottdavis:
how does this take in to account that past versions of browsers may still be in use? not everyone lives on the bleeding edge.

chriseppstein:
This patch is kind of ridiculous. I appreciate the intent, but I cannot merge to stable a patch that drops support for a half-dozen mixins and changes the api of others. I'd be happy to accept a patch to stable that maintains our current API but reduces the generated output to only what is necessary. On master we can also accept a patch that brings the border-radius API more in line with the CSS property and responsibly deprecates the existing API.

PR #447, opened 2011-06-29
Removing -ms-filter as it drives IE9+ crazy. plus existing filter works a

nimbupani:
Removing -ms-filter as it drives IE9+ crazy. plus existing filter works anyway for IE9-

More documented here: https://github.com/paulirish/html5-boilerplate/pull/354 (also note http://css3please.com no longer serves `-ms-filter` with `filter`).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -7,7 +7,7 @@ GIT
 PATH
   remote: .
   specs:
-    compass (0.11.3.b352e8b)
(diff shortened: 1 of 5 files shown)


mathiasbynens:
FWIW the relevant css3please.com commit is this one: https://github.com/paulirish/css3please/commit/403adb23aad8100f6b3ea111d18dc4e694cb28df Haven’t had a problem with it since.

PR #451, opened 2011-07-01
Fix absolute path detection on Windows

schniper:


chriseppstein:
thanks for the patch.

Issue #449, opened 2011-06-30
all-*-sprites  should accept and honor offsets


caged:
The `sprite` mixin allows you to supply and x and y offset.  `all-*-sprites` uses the `sprite` mixin internally, but doesn't allow you to specify offsets.

And, a closely related issue.  If you'd like me to break these into two tickets, let me know.
Offsets are not honored when compass generates :active, :hover and :target selectors from _active.png, _hover.png and _target.png images.  Compass will also generate excessive selectors.  

See https://gist.github.com/1057445 for an example of both issues.

Edit: I'm using the rails31 branch

scottdavis:
this is a bug bug what do you mean by "excessive selectors"?

caged:
This is the output I get from a single _active.png image. Is there a compelling reason to generate all three of these?

`.icons-demographics:active, .icons-demographics.demographics_active, .icons-demographics.demographics-active`

scottdavis:
yes we decided to do this incase you wanted to use a state else were all you have todo is use @extend. instead of booting the sprite engine to figure out the already known location of something and repeating yourself

scottdavis:
fixed

PR #453, opened 2011-07-02
Sprite offset fix

caged:
Fix for #449


caged:
Not sure what happened here.  This should be applied to the rails31 branch, not stable.

Issue #399, opened 2011-05-24
Display Helper: elements-of-type(html5) missing elements


adion:
I believe the `html5` display value is missing a few tags: http://web-design.blogs.webucator.com/2010/10/14/list-of-new-html5-tags/

Also, `dialog` isn't a valid HTML element.

chriseppstein:
The list from that site is:

```
article, aside, audio, canvas, command, datalist, details, embed, figcaption, figure,
footer, header, hgroup, keygen, mark, meter, nav, output, progress, rp, rt, ruby,
section, source, summary, time, video, wbr
```

I think some of these are inline elements so we need to sort out which is which and then change blueprint to only reset html5 block elements.

It would be great if someone could make this patch.

dira:
Started working on the patch.

chriseppstein:
Fixed by 07819ccfbdcf3055d53877aa93187ce5f50df8ff

Issue #339, opened 2011-04-14
filter-gradient requires utilities/general/hacks


joliss:
Using filter-gradient throws the following error message:

```
Syntax error: Undefined mixin 'has-layout'.
        on line 77 of /var/lib/gems/1.9.1/gems/compass-0.11.beta.6/frameworks/compass/stylesheets/compass/css3/_images.scss, in `has-layout'
        from line 77 of /var/lib/gems/1.9.1/gems/compass-0.11.beta.6/frameworks/compass/stylesheets/compass/css3/_images.scss, in `filter-gradient'
```

It goes away if I put `@import "compass/utilities/general/hacks";` at the top.  Shouldn't filter-gradient import that module automatically?

(I'd send you a pull request, but I'm not sure where to best put the @import.)

joliss:
Also: I've only used it very briefly, but so far, Compass seems most awesome!  Thanks for all the hard work!

chriseppstein:
Yes. Thanks for the bug, we will add the import.

PR #462, opened 2011-07-06
Reset only the block HTML elements #339

dira:
[#339](https://github.com/chriseppstein/compass/issues/399)
- used elements-of-type(html5-block) in all the reset stylesheets
  (instead of duplicate tag enumerations)
- removed dialog
- cavas is inline, not block


chriseppstein:
This was merged to stable as 07819ccfbdcf3055d53877aa93187ce5f50df8ff. Thanks!

PR #475, opened 2011-07-12
Sort sprites in map by file size

ai:
Currently sprites in map sort alphabetically. For most developers it is mean randomly. I think it will be better to sort sprites by file sizes, becuase then half of sprites will be loaded faster. Now we have some bad examples, for example http://static.groupon.ru/images/jobs-sdf196c2b3a.png very big background is in first position, but it only for decoration.


chriseppstein:
@jetviper21 thoughts?

scottdavis:
"half of sprites will be loaded faster" im curious to how true this is. The question i have it it necessary to features like this. The issue i have with this is with the bin packing method that is in the works sorts by width or height depending on layout. this is going to add a lot more "what if's" to the current roadmap. If you can prove that it truly does effect render time then im game but i think your going to find that its negligible unless your sprite files are enormous and in that case you may not want to sprite.

scottdavis:
my main concern about spiriting large images is mostly browser memory usage.

scottdavis:
see http://blog.vlad1.com/2009/06/22/to-sprite-or-not-to-sprite/ 3rd paragraph

ai:
It will be helpful not only for big files. For example, in flags test files this patch put simple flags to top of file and half of flags will be loaded fatser too.

ai:
Also if I remove big background from my sprite they will be contain very small and middle image. And load small icons faster will be better.

ai:
Also, because small images will be together with this patch, it will join free area in right part of sprite map and reduce file image (of course, different will be small, but it we don’t lose anything).

scottdavis:
im not against this but having to support 3+ ways of sprite layout is going to get to be a hassle thats why im being resistant. if this is needed maybe we should work out a way that people can extend the sprites to there liking?

chriseppstein:
We'll have bin pack, horizontal pack, and vertical pack soon. This sounds like it's a bin pack competitor, yes? These three packs seem like enough to me but maybe I'm missing some finer point.

scottdavis:
well horizontal pack actually solves this same problem

ai:
I just think that file size sorting is better, that sort alphabetically. So I didn’t push another layout method, just try to make better current method.

scottdavis:
So the other week I implemented horizontal packing which will solve your load smaller images first problem. So I think this implementation is unneeded moving forward https://github.com/chriseppstein/compass/commit/18c79f51a24a60e69bfce94b258cfaa723fb90b7

ai:
Why horizontal packing solve my problem? After half size user will not see any full loaded images :).

scottdavis:
Touché! If you apply the to master for vertical packing only I'll accept it

ai:
Horizontal packing is in better_sprite_packing branch? Ok, I will create another pull request. Also, I think will be better to sort files in horizontal packing by image height, because it will join free bottom spaces.

scottdavis:
Master

scottdavis:
https://github.com/chriseppstein/compass/commit/b581d2e1a7720df5e812730a3232c4834b409498#L1R19

PR #485, opened 2011-07-20
trivial documentation change

adamnbowen:
"to the right" doesn't make sense with how the page is laid out 

"listed below the display value" is a bit more relevant to the current documentation layout


MarioRicalde:
http://compass-style.org/reference/compass/helpers/display/

Agreed. It was confusing when I saw that page. I think below makes more sense :)

PR #495, opened 2011-07-27
Rails integration fix

joerixaop:
Currently setting the http_path to "/something" or even "/something/" in a rails project results in a default http_images_path of "/somethingimages". This patch changes that to "/something/images"

### DIFF
--- a/lib/compass/app_integration/rails/configuration_defaults.rb
+++ b/lib/compass/app_integration/rails/configuration_defaults.rb
@@ -36,19 +36,21 @@ module Compass
         end
 
         def default_http_images_path
-          "#{top_level.http_path}images"
+          # Relies on the fact that this will be loaded after the "normal"
+          # defaults, so that method_missing finds http_root_relative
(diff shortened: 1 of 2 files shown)


scottdavis:
just to be the devils advocate here setting http_path to "/something/" doesn't solve this?

joerixaop:
No, apparently the trailing "/" is removed from the http_path, either way
the behaviour without the rails integration it does work without the
trailing slash in http_path, so I think that purely for consistency the
behaviour should be changed even if it worked with "/something/"

scottdavis:
can you make this against master? i don't like merging things into stable

joerixaop:
Sure I didn't even notice that I made the pull request against "stable", after all the last commit on this branch is a pull request from only a day ago (admittedly only a difference in wording).

When running the tests against the master branch I noticed many (29) errors with encoding mismatching that weren't there in stable (tests run with 1.9.2 in both cases)

PR #509, opened 2011-08-09
user-select mixin with docs and tests

mirisuzanne:
The only problem here is I haven't got the compass docs or tests running on lion yet, so those aren't tested properly. They look right to me. Anyone care to verify and pull?

### DIFF
--- a/doc-src/content/CHANGELOG.markdown
+++ b/doc-src/content/CHANGELOG.markdown
@@ -14,6 +14,11 @@ The Documentation for the [latest stable release](http://compass-style.org/docs/
 
 The Documentation for the [latest preview release](http://beta.compass-style.org/)
 
+0.11.6 (UNRELEASED)
+-------------------
+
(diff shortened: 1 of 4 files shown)


mirisuzanne:
all tests pass.

chriseppstein:
@ericam please add a changelog entry and merge it in :)

rykov:
According to [MDN](https://developer.mozilla.org/en/CSS/-moz-user-select), the possible values for _-moz-user-select_ are none, all, text, and -moz-none.  When I do _-moz-user-select: -moz-text_, it is ignored.

The following implementation has worked for me:

```
@mixin user-select($select) {
  $select: unquote($select);
  @include experimental(user-select, $select,
    -moz, -webkit, not -o, not -ms, -khtml, official
  );
}
```

mirisuzanne:
care to submit a patch?

rykov:
Done: chriseppstein/compass#581

mirisuzanne:
merged. thanks!

korny:
I noticed I still have to manually require this mixin in 0.11.7. Shouldn't compass/css3/user-interface be included by compass/css3?

Issue #477, opened 2011-07-15
inline-font-files is not working


rolang:
Inline-font-files breaks the CSS. It seems like the encoding is wrong. I used UTF-8 and the base64 string seem strange. Like when you open a Word file with notepad.
Tested with sass version 3.1.4, compass version 0.11.5, on windows 7 and opensuse 11.3, with font-files ttf, eot and woff. Can't copy and paste the string here, it won't work.

tyrinslys:
I see the same problem, and in addition the closing characters for the @font-face are not printed, so it breaks css for the rest of the file.

<pre>
$fontOstrich: "ostrich-bold.ttf";
@include font-face("ostrich", inline-font-files($fontOstrich, truetype));

yeilds:
@font-face {
font-family: "ostrich";
src: url('data:font/truetype;base64, [lots of font data here]
</pre>

expected to see these ending characters, but they were not present.

<pre>
')}
</pre>

botandrose:
Turns out the font file wasn't being base64 encoded. I've sumitted a pull request to fix this: https://github.com/chriseppstein/compass/pull/520

chriseppstein:
that landed a while ago. closing.

PR #520, opened 2011-08-25
Inline fonts weren't being base64 encoded.

botandrose:
Should close issue #477.

We might consider renaming the inline_image_string function to something more generic.


scottdavis:
pull requests need to be made against master if you wouldn't mind redoing this.

botandrose:
Done.

PR #535, opened 2011-09-05
3D Browsers list should be configurable

ZeeAgency:
Hi,

I heard that Firefox will soon support 3d transforms and one day IE & Opera also will. So instead of having 3d supporting vendors prefixes inside mixins I added them as a variable and used it to apply-origine, transform, perspective, etc...

I'm sorry I don't have much time for testing but so far it works great for me.

Thanks

Julien

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_transform.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_transform.scss
@@ -88,8 +88,19 @@ $default-skew-x      : 5deg                  !default;
 // The default y-angle for skewing
 $default-skew-y      : 5deg                  !default;
 
+// The default 2D supporting browsers list
+$default-support2d-moz: true;
+$default-support2d-webkit: true;
+$default-support2d-o: true;
+$default-support2d-ms: true;
+$default-support2d-khtml: false;
+
 // The default 3D supporting browsers list
-$default-supporting3d-browsers: -moz, -webkit, -o, -ms, not -khtml, official;
+$default-support3d-moz: false;
+$default-support3d-webkit: true;
+$default-support3d-o: false;
+$default-support3d-ms: false;
+$default-support3d-khtml: false;
 
 
 // **Transform-origin**
@@ -107,11 +118,11 @@ $default-supporting3d-browsers: -moz, -webkit, -o, -ms, not -khtml, official;
   $only3d: $only3d or -compass-list-size(-compass-list($origin)) > 2;
   @if $only3d {
     @include experimental(transform-origin, $origin,
-      $default-supporting3d-browsers
+      $default-support3d-moz, $default-support3d-webkit, $default-support3d-o, $default-support3d-ms, $default-support3d-khtml, official
     );
   } @else {
     @include experimental(transform-origin, $origin,
-      -moz, -webkit, -o, -ms, not -khtml, official
+      $default-support2d-moz, $default-support2d-webkit, $default-support2d-o, $default-support2d-ms, $default-support2d-khtml, official
     );
   }
 }
@@ -150,11 +161,11 @@ $default-supporting3d-browsers: -moz, -webkit, -o, -ms, not -khtml, official;
 ) {
   @if $only3d {
     @include experimental(transform, $transform,
-      $default-supporting3d-browsers
+      $default-support3d-moz, $default-support3d-webkit, $default-support3d-o, $default-support3d-ms, $default-support3d-khtml, official
     );
   } @else {
     @include experimental(transform, $transform,
-      -moz, -webkit, -o, -ms, not -khtml, official
+      $default-support2d-moz, $default-support2d-webkit, $default-support2d-o, $default-support2d-ms, $default-support2d-khtml, official
     );
   }
 }
@@ -182,7 +193,7 @@ $default-supporting3d-browsers: -moz, -webkit, -o, -ms, not -khtml, official;
 // values from 500 to 1000 are more-or-less "normal" - a good starting-point.
 @mixin perspective($p) {
   @include experimental(perspective, $p,
-    $default-supporting3d-browsers
+    $default-support3d-moz, $default-support3d-webkit, $default-support3d-o, $default-support3d-ms, $default-support3d-khtml, official
   );
 }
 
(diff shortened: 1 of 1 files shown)


scottdavis:
can you create a test for this? im ok with merging this but I hate merging things without test cases

ZeeAgency:
Ok but could you just tell me how should I test for this ? As I'm just updating existing code, maybe I could use some existing tests for transforms mixins ?

scottdavis:
is the output css changing?

ZeeAgency:
Yes... For instance :

@include translate3d(0, 0, 0);

Now (no setting)
-webkit-transform: translate3d(0, 0, 0);
transform: translate3d(0, 0, 0);

After (setting : $default-supporting3d-browsers: -moz, -webkit, -o, -ms, not -khtml, official;)
-moz-transform: translate3d(0, 0, 0);
-webkit-transform: translate3d(0, 0, 0);
-o-transform: translate3d(0, 0, 0);
-ms-transform: translate3d(0, 0, 0);
transform: translate3d(0, 0, 0);

I can also push with default settings like the current behaviour : $default-supporting3d-browsers: not -moz, -webkit, not -o, not -ms, not -khtml, official, so the output won't change by default but if people want to change this, then, now it's a setting.

scottdavis:
https://github.com/chriseppstein/compass/blob/stable/test/fixtures/stylesheets/compass/sass/transform.scss and
https://github.com/chriseppstein/compass/blob/stable/test/fixtures/stylesheets/compass/css/transform.css

need to be updated accordingly

ZeeAgency:
Ok but what if I'm pushing with default settings as current behaviour ? Any SASS or any CSS output won't change at all.

chriseppstein:
I like what this patch is trying to do, but it's not acceptable in its current state. The changes are functionally broken given how your passing around the list of acceptable browsers.

ZeeAgency:
Hi Chris,

"Not to be rude, but did you actually test this patch?"
Yup, I'm sorry : I spoke too fast, I tested the wrong file...

You're right, this implementation was totally wrong.
I made another commit which should fix this and also offer 2D support configuration, may you have a look ?
https://github.com/ZeeAgency/compass/commit/1861e7c1d2e1aafa0677cc5be23860c1b70d5508

About -ms (or -o) and 3D support : Like I said, it will one day or the other. IMO, it's not Compass's job to tell which browsers supports 3D or not because it's like browser sniffing and it means we should update all our CSS everytime a new browser supports 3D, exactly what we should never do. I'm already using Modernizr today to detect this kind of features and if FF or any browsers supports it tomorrow, no CSS would be targeting it. 

For instance :
```#foo {
    // No transforms
    html.no-csstransforms.no-csstransforms3d & {
        left: 10px;
    }
    // 2D
    html.csstransforms.no-csstransforms3d & {
        @include translateX(10px);
    }
    // 3D
    html.csstransforms3d & {
        @include translate3d(10px, 0, 0);
    }  
}

```
This code works fine today on FF but it won't tomorrow...

On the other hand, we should absolutely keep this configurable because of some projects where we don't need to target every browsers : for instance, I'm currently working on a webkit-only web app so I don't need extra weight from other vendor prefixes, even for 2D.
```

chriseppstein:
I'm -1 on enabling any support that does not yet exist, especially if it's configurable.

I agree that modernizr is a great tool. I use it on all my apps. But I'm not going to make it hard dependency for compass projects. The evolution of browsers is not so fast that we can't keep up as a community. If it comes to it, we can even write tests that use modernizr to check our behavior.

ZeeAgency:
I'm not saying Modernizr should be a dependency, my only point is "this should be a setting". If you look at my commit, I put the same settings as before : only webkit prefix for 3D.

ZeeAgency:
What do you mean ? Changing @mixin experimental or what ?

chriseppstein:
@ZeeAgency What I mean is to have a single list and test if the list includes some value. Unfortunately this is harder than it should be and we should make it easier in sass. I have this on a lit of things to do in sass and then I will make this change. I'm going to close this request. Until then we can update the call to the `experimental` method as browser support changes. Thanks for your patch, it's certainly the right direction for compass to go.

ZeeAgency:
@chriseppstein

Time to up this thread because :
1. It's a bit old now
2. No solution came, afaik Sass is still missing some features in order to do this well 
3. I actually came to think this is the wrong way to go...

**TL;DR**
We should **only** rely on `$experimental-support-for-vendor` variables.

I'm not saying this only because of the recent -webkit- prefix [drama](http://www.glazman.org/weblog/dotclear/index.php?post/2012/02/09/CALL-FOR-ACTION%3A-THE-OPEN-WEB-NEEDS-YOU-NOW) [episode](http://css-tricks.com/tldr-on-vendor-prefix-drama/), my only point is : this way of doing is **not** future-proof. It requires us, developers, to update Compass and all of our stylesheets _everytime_ a new browser supports 3d, transitions and probably other features I'm missing.

Look... [This file](https://github.com/chriseppstein/compass/blob/stable/frameworks/compass/stylesheets/compass/css3/_transform.scss#L5) didn't change in almost **a year** !

> Safari is the only browser that currently supports 3D transforms."

This is [not true](http://caniuse.com/#search=transform) anymore, and this will become falser months after months, this is clearly not a maintainable approach.

In my humble opinion, Sass & Compass are great tools in syntax enhancing, authoring, reusability and many other things but they shouldn't care that much about browsers panorama. I have two exceptions in mind : handling inconsistencies (webkit gradients for instance) and only prefix the needed properties/values regarding of browser implementation. I'll clear the last point if needed : if a browser decided to implement an experimental feature without prefixing, Compass shouldn't have to, if a browser **didn't** implement an experimental feature yet, Compass **should** prefix it ! Future-proof, QED.

Let just developers take the responsibility for this, we have feature detection ! I don't mean Modernizr should be a requirement for Compass, I mean Modernizr (or other feature detection libs) **is** a requirement for people wild enough to deal with experimental features. 

The sad story is today, using Compass + Modernizr will hurt your feelings : I had many _compassed_ sites where 2d transforms stopped to work because of Firefox supporting 3d and conditional styling through Modernizr failed because -moz- was missing in .csstransforms3d case...

The same way we can say CSS preprocessors are not creating CSS bloat, devs are, we can say Compass will not taking care about this, devs will. Again, we have the tools for this.

I'm [quoting you](http://lists.w3.org/Archives/Public/www-style/2012Feb/1009.html), Chris, here : 

> We have the luxury of being able to provide updates, deprecate syntax and introduce new concepts at a much faster rate than the CSSWG.

So first, we have the luxury to warn people about this (radical ?) change I'm proposing, that's a good one. People can tell what they think about it, much faster than the CSSWG, again, and they can be warned when they'll update Compass. No hurt.

Next, without a doubt, Compass & Sass are progressing faster than the CSSWG, but I don't think it can move as fast as browser's updates. Browsers are committing new versions at an impressing rate nowadays. And even if it can, again, it would require developers to update their code each time. That's not an option, right ?

Thanks in advance for your feedback on this,

Julien

PS : Actually, there's [another option](http://leaverou.github.com/prefixfree/) but I would personally rather use it as a spare wheel, in addition to Compass.

PR #537, opened 2011-09-06
Fixed error when trying to compile examples by rake command

pzgz:
Getting errors when running "rake examples", made following updates:
- Use https instead http for github access 
- Set use SSL flag for http client 
- Fix the link for compass-960-plugin


pzgz:
Also extend the http connection open timeout a little bit to avoid time out issue in slow connectivity.

PR #538, opened 2011-09-06
Inline fonts weren't being base64 encoded.

botandrose:
Should close issue #477.

We might consider renaming the inline_image_string function to something more generic.


scottdavis:
mind creating a test?

botandrose:
That was my first thought, but it appears that this entire section of compass doesn't have any test coverage, so I was a bit reticent to start a whole new test section. Where do you think we should add this test? /test/units, probably?

Correct me if I'm wrong about this, and just totally missed the tests!

scottdavis:
test/units/sass_extensions_test.rb

botandrose:
Not sure if this is the best approach, but it's correctly verifying the desired behavior, at least. What do you think?

scottdavis:
looks good to me, il get this merged soon. I only push for tests because we are lacking them in a lot of areas

botandrose:
Roger that. I was a bit surprised by the lack of coverage, considering how nice a library Compass is! Thanks for pushing for the test.

Issue #544, opened 2011-09-09
font_files helper requires font types to be named explicitly


ravinggenius:
Is this by design? I'd like to be able to leave off the font type if it could be inferred by the extension. I don't mind refactoring this myself. Any comments?

ravinggenius:
I went ahead and made the changes. Feel free to adjust if necessary.

scottdavis:
looks fine to me

ravinggenius:
I added a pull request (issue #545).

Issue #543, opened 2011-09-09
font_files helper has no test coverage


ravinggenius:
font_files helper really should have test coverage. I'd like to refactor it to be less obtuse, but it needs tests for the current implementation first.

ravinggenius:
I added a pull request (issue #545).

PR #549, opened 2011-09-12
Test and refactoring of font_files (issues #543 and #544) on master branch

ravinggenius:


chriseppstein:
looks good. thanks for the patch.

PR #550, opened 2011-09-12
Clean up Compass::SpriteImporter.

nex3:
The main functional changes here are:

1) Making the :filename option on the generated CSS files of a format that is
   acceptable for SpriteImporter#find.
2) Making #find_relative return nil, since there are never relative imports
   within a generated stylesheet.

This change also brings the structure of SpriteImporter more inline with its
intended use. Instance variables such as @url and @options don't make sense,
since the same instance is used to import every sprite stylesheet; now these
variables are local and passed from function to function.

This change (specifically functional change 1 above) should also fix issue
sass#114.

### DIFF
--- a/lib/compass/commands/sprite.rb
+++ b/lib/compass/commands/sprite.rb
@@ -39,10 +39,11 @@ module Compass
 
       def perform
         relative_uri = options[:uri].gsub(/^#{Compass.configuration.images_dir}\//, '')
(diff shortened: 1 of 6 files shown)


chriseppstein:
@scottdavis is this bug fixed on master?

scottdavis:
Yes

PR #551, opened 2011-09-13
The return value assignments are reversed in this case

nbarthelemy:
the name and path variables needed to be reversed.


scottdavis:
does this break any tests?

nbarthelemy:
No because there are no tests that test if nested directories work. ( i.e. @include 'ui/sprites/button/*.png' ) All of the tests use a single level deep, so the name and path would be the same.

scottdavis:
hmm i though i had a test for nested

scottdavis:
your right let me make a failing test case then ill apply this

PR #554, opened 2011-09-16
fixes a bug on Windows systems

schmittjoh:
There might be a better way to normalize file paths, but I don't know the code base well enough to propose it. 

This basically fixes cases where paths such as "C:\Users\Johannes" are compared vs "C:/Users/Johannes" which incorrectly results in the error being raised.

### DIFF
--- a/lib/compass/compiler.rb
+++ b/lib/compass/compiler.rb
@@ -32,7 +32,7 @@ module Compass
     end
 
     def stylesheet_name(sass_file)
-      if sass_file.index(from) == 0
+      if sass_file.gsub(/\\/, '/').index(from) == 0
         sass_file[(from.length + 1)..-6]
       else
         raise Compass::Error, "You must compile individual stylesheets from the project directory."


schmittjoh:
What do you think about only doing the replacement when running on Windows?

scottdavis:
basically we should be using File.join for everything that should take care of the `\` on windows might be time for some refactoring

chriseppstein:
Closing this pull request. Need to find a better solution.

Abhoryo:
Hi @chriseppstein, did you find a better solution for this bug ?

scottdavis:
not yet neither of use are running windows so we need someone on windows and ruby fu to work on this

Abhoryo:
```
"C:\\Ruby193\\bin\\compass.BAT" "compile" 
"C:\\Users\\Symfony\\AppData\\Local\\Temp" 
"--images-dir" "C:\\Symfony\\app/../web/images" 
"--config" "C:\\Users\\Symfony\\AppData\\Local\\Temp\\assC503.tmp" 
"--sass-dir" "" "--css-dir" "" 
"C:/Users/Symfony/AppData/Local/Temp/assC514.tmp.scss"
```

I don't understand why 3 of 4 paths are correctly managed and not the last one. It doesn't have the same code as the other?

Issue #527, opened 2011-09-01
undefined method `static_root' for #<Sprockets::Index>


rwz:
I've made a blank new rails 3.1 app with compass and one sass file with sprites. And the `assets:precompile` rake task fails on this setup with two different error messages.

You can see the test app here: https://github.com/rwz/compass_static_root_bug

guilleiguaran:
The error of the second run was fixed with #2768, I don't have idea about the first one

guilleiguaran:
The first one:

https://github.com/chriseppstein/compass/commit/7147147e24121ffb03044c638a6a8b02b8fccb7f#L0R68

static_root was deprecated in Sprockets

kremso:
@guilleiguaran: I was able to fix the first error by using the code you suggested.

https://github.com/kremso/compass/commit/1e5debe9388505d9dc993d4599d8dca696833f37

However, that didn't fix the second error. I was able to workaround that one by setting this in production.rb

   config.assets.compile = true

guilleiguaran:
@kremso, try using rails from 3-1-stable branch

chriseppstein:
confirmed. this is a version inconsistency between rails and sprockets. nothing to do w/ compass.

rwz:
What about the static root bug? It's still not fixed in current master.

chriseppstein:
my bad. Compass does refer to static root.

adamhooper:
I've switched to kremso's commit 1e5debe9388505d9dc993d4599d8dca696833f37 instead of 0.12.alpha.0. It works for me. Any reason not to merge it?

guilleiguaran:
@adamh looks like @kremso haven't send a pull request

rwz:
Switched on rails 3.1.1.rc1 with kremso's compass fork. Works fine so far.

kremso:
Pull request sent.

matsimitsu:
Any updates on this one? Still getting the issue :)

scottdavis:
this should be in master https://github.com/chriseppstein/compass/commit/fe4edff7d4980b50ae377336cf8ccfbc284bfaaa

agibralter:
Any chance for a 0.12.alpha.1?

PR #556, opened 2011-09-17
Remove call to deprecated static_root

kremso:
This is a patch for issue #527


kevmoo:
This patch looks like it was taken.

Should this be closed?

fe4edff7d4980b50ae377336cf8ccfbc284bfaaa

PR #564, opened 2011-09-21
fixed limitation of last-child selector (not needed anymore ;) )

brikou:
- I changed padding to margin if user want to add background color fe
- last-child / first-child psuedo selector are no more needed because I added a negative leading margin


mirisuzanne:
I agree with the switch to margins, but disagree with the other. 

This is better in some use cases, but it introduces unexpected behavior. The container won't be the size or position that I expected. If I add borders, they won't be where I expected. Etc. All to hack around a CSS selector that is now implemented in all modern browsers?

More reasonable would be to switch so only :first-child is needed, as first-child has support all the way back through IE7.

chriseppstein:
Thanks for the patch. I agree with @ericam. Please submit a new PR with the `:first-child` approach if you're so inclined.

PR #572, opened 2011-09-26
Added `sprite-replace-text` mixin

ultimatedelman:
Added `sprite-replace-text` mixin for replicating the `replace-text-with-dimensions` mixin using sprites.

### DIFF
--- a/frameworks/compass/stylesheets/compass/utilities/sprites/_base.scss
+++ b/frameworks/compass/stylesheets/compass/utilities/sprites/_base.scss
@@ -72,7 +72,6 @@ $disable-magic-sprite-selectors:false !default;
 @mixin sprite-replace-text ($map, $img-name){    
     @include hide-text;
     background: sprite($map, $img-name) no-repeat;
-    display: block;
     width: image-width(sprite-file($map, $img-name));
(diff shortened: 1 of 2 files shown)


scottdavis:
not to be a grump but ill merge this once there is a test case and documentation if you have issues let me know

ultimatedelman:
Sorry for my ignorance... this is my first github public repo commit. How would you like me to provide a test case and/or more documentation? I put in some comments above the mixin in my commit documenting how to use it. What else should I do?

scottdavis:
oh, sorry didn't mean to have a negative tone. um for starters the test files for sprites are in test/integrations/sprite_test.rb and actually don't worry about the docs what you did should be ok i had a brain fart.

scottdavis:
basically in the test you need to verify that the correct css gets generated if you get it close enough i can get it working correctly it sometimes has white space issues

ultimatedelman:
Scott - 

I added a test, but I've admittedly never programmed in ruby before, so it probably doesn't work! Please let me know if there's anything else I can do for you. Also, if you do get it working, please tell me what you did so I can learn :)

scottdavis:
ok i added this double check the commit and make sure everything looks right

PR #573, opened 2011-09-27
add group parameter to rails initializer to run on asset precompilation

saks:
Hi there, asset precompilation task was [changed](https://github.com/rails/rails/commit/b4798894d35e2f1fbd53aad8cd230de154650339) in rails 3-1-stable branch. There is commit message:

To the app developer, this means configuration add in
config/initializers/\* will not be executed.

Plugins developers need to special case their initializers that are
meant to be run in the assets group by adding :group => :assets.

I have added this option.

### DIFF
--- a/lib/compass/app_integration/rails/actionpack3/railtie.rb
+++ b/lib/compass/app_integration/rails/actionpack3/railtie.rb
@@ -37,7 +37,7 @@ end
 
 module Compass
   class Railtie < Rails::Railtie
-    initializer "compass.initialize_rails" do |app|
+    initializer "compass.initialize_rails", :group => :assets do |app|
       # Configure compass for use within rails, and provide the project configuration
       # that came via the rails boot process.
       Compass::AppIntegration::Rails.initialize!(app.config.compass)


avocade:
Is this what's causing @import "compass"; or @import "susy"; to fail during assets:precompile?

Works great having them in development, but when precompiling it seems compass and susy aren't required properly.

```
group :assets do
  gem 'sass-rails', :git => 'git://github.com/rails/sass-rails.git', :branch=>'3-1-stable'
  gem 'coffee-rails', '~> 3.1.0'
  gem 'uglifier'
  gem 'compass', '~> 0.12.alpha'
  gem 'compass-susy-plugin', :require => 'susy'  # http://susy.oddbird.net/tutorial/
  gem 'oily_png' # for faster compass sprite PNG compilation 
end 
```

saks:
it's about import of compass while precompilation

scottdavis:
This patch needs to be made against master your current fix is applying to the rails 3.0.x railtie and not the 3.1.x railtie.

PR #591, opened 2011-10-06
more attractive whitespace for prefixed and aligned styles

paulirish:
Here we take these aligned properties and values and order their prefixes to create a very nice visual waterfall.

So it's not only easy to look at as a group (as they are aligned), but attractive as well.

This approach is used on http://css3please.com/ and written by desandro at http://dropshado.ws/post/2054719546/css-formatting

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_shared.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_shared.scss
@@ -5,34 +5,34 @@
 // the implementations are identical except for the property
 // prefix.
 @mixin experimental($property, $value,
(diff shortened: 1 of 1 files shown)


scottdavis:
Looks good ill cherry pick this over to master if you plan on making anymore pull requests can you do them against master? I don't like updating stable unless we do a release

paulirish:
No problemo. Chris pointed me to the stable branch so I grabbed from there. 
I'll work on master from now on, thx scott.

scottdavis:
np thanks for the pull requests!

PR #595, opened 2011-10-08
Optimizations for sprites

brainopia:
Hello, I've been profiling performance of sprite generation and noticed a couple of places which can be improved.

First of all, SpriteMap doesn't reset stored `mtime` of sprite after regeneration so not only first call to old sprite invokes regeneration but all subsequent ones too.

Secondly, `stat` call is issued for every sprite image when a few can be enough to see that sprite is outdated.

I've fixed aforementioned issues and removed trailing whitespace too :)


scottdavis:
Thanks, we know sprites are slow and they always will be. you  can gain a huge performance boosts by using the rmagick engine instead of chunky png if its becoming a real problem. Ill merge these in soon.

Issue #526, opened 2011-08-31
0.12.alpha.0 and "File to import not found..."


scottkf:
``` ruby
ActionView::Template::Error (File to import not found or unreadable: blueprint/grid.
```

It's actually true for every file I try to import. Solving this was simply adding

``` ruby
require 'compass'
require 'sass-rails'
```

to `application.rb`

However, is this something that should be done automatically, or, what am I missing here?

scottdavis:
run `bundle exec compass frameworks` in curious what the output is

bmaland:
I suspect this was due to only including the compass and sass-rails gems in the asset group in Gemfile, they have to be in development as well.

steverandy:
I was having the same issue with Rails 3.1.0, but it was ok when I upgraded to Rails 3.1.1.rc1.

But now I’m getting the same error with Rails 3.1.1.rc2.

rwz:
Same here. Worked on rc1, fails on rc2.

```
% bundle exec compass frameworks
Available Frameworks & Patterns:

  * blueprint
    - blueprint/basic      - A basic blueprint install that mimics the actual blueprint css.
    - blueprint/buttons    - Button Plugin
    - blueprint/link_icons - Icons for common types of links
    - blueprint/project    - The blueprint framework.
    - blueprint/semantic   - The blueprint framework for use with semantic markup.
  * compass
    - compass/ellipsis     - Plugin for cross-browser ellipsis truncated text.
    - compass/extension    - Generate a compass extension.
    - compass/pie          - Integration with http://css3pie.com/
    - compass/project      - The default project layout.
```

PR #603, opened 2011-10-12
Fix Railtie initialization when initialize_on_precompile is disabled

jdelStrother:
Heya,
Rails 3.1.1 adds config.assets.initialize_on_precompile, which can be set to false to avoid initializing the entire application environment.

This commit tells Rails that it's necessary to initialize compass even when initialize_on_precompile is false.

Any thoughts?

### DIFF
--- a/lib/compass/app_integration/rails/actionpack31/railtie.rb
+++ b/lib/compass/app_integration/rails/actionpack31/railtie.rb
@@ -76,7 +76,7 @@ end
 
 module Compass
   class Railtie < Rails::Railtie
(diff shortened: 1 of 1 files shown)


msaffitz:
+1

Thanks-- this works for me on 3.1.1 final.  Note that this also likely resolves #526

scottdavis:
does this work on 3.1.0 also?

jdelStrother:
Yep - unrecognised options to the rails Initializers are just ignored.

Issue #599, opened 2011-10-11
underline.png in vertical_rhythm is missing


jonassmedegaard:
The mixin debug-vertical-alignment by default use underline.png but that file is nowhere to be found in the source of Compass.

scottdavis:
@ericam do you have the png file hiding someplace?

mirisuzanne:
Nope. Sorry. Don't know where that came from or where it went to. 

But it wouldn't be very useful if it was there, because it needs to match your actual site settings. You really should be making your own or using the new grid background module. Compass does include a simple way to make your own in the command line: 

```
compass grid-img W+GxH [path/to/grid.png]
```

(Width + Gutter x Height)

I'd recommend changing debug-vertical-alignment so it becomes a shortcut for baseline-grid-background, but allowing both to take an (undefined by default) image argument as fallback for older browsers. I'd also add a note in the docs pointing to the grid-img command line tool. Not sure how soon I can get to it, but if someone wants to submit a patch I'll be happy to look it over.

scottdavis:
ok ill move this to the 0.12.1 milstone

barraponto:
@ericam can't we use the `on_style_saved` callback to generate this file? Trying to use it out of the box and missing an image file is frustrating.

scottdavis:
I think what he was getting at is everyone's underline.png is different

mirisuzanne:
@barraponto It is certainly a problem for things not to work out of the box. But right now we have a bug (missing file) and a broken approach (an image that should be different for everyone). I'm suggesting we fix the approach, not just the bug.

barraponto:
@ericam well, on_style_saved callbacks would fix it while we change the approach. If I knew how to get a Compass plugin to modify the values in config.rb, I'd write a patch.

BTW, I believe we can try generating the grid with repeating gradients, what do you think?

mirisuzanne:
We already are using gradients for the grid-background mixins that I'm proposing we use here. We aren't using repeating gradients because they can't account for the concept of a side-gutter. Though we could use some logic to create simpler repeating gradients (less code) when there are no side-gutters (or they are equal to the internal gutters). 

Compass doesn't actually have built in support for repeating gradients yet, right? Or did I miss something?

barraponto:
@ericam no support for repeating gradients, although @chriseppstein did post a workaround on #401

mirisuzanne:
https://github.com/chriseppstein/compass/pull/615

PR #615, opened 2011-11-10
Stable

mirisuzanne:
addresses issue 599: https://github.com/chriseppstein/compass/issues/599


mirisuzanne:
how did those other commits get mixed in here?

Issue #296, opened 2011-03-08
Bug with repeat-x


aurelien-baudet:
Hi,

I have a sprite with many images and not with the same size. The width of the generated sprite is 1024px. I have a 22px image that I want to put in this sprite and repeat it horizontally. I used the variable
    $<map>-<sprite>-repeat: repeat-x;
as described in the documentation but I got the error:
    ChunkyPNG::OutOfBounds on line ["233"] of D: Background image width is too small!

stack:

<pre>
D:/Applications/developpement/Ruby/lib/ruby/gems/1.9.1/gems/chunky_png-0.12.0/lib/chunky_png/canvas/operations.rb:48:in `replace'
D:/Applications/developpement/Ruby/lib/ruby/gems/1.9.1/gems/compass-0.11.beta.2/lib/compass/sass_extensions/functions/sprites.rb:153:in `block in construct_sprite'
D:/Applications/developpement/Ruby/lib/ruby/gems/1.9.1/gems/compass-0.11.beta.2/lib/compass/sass_extensions/functions/sprites.rb:146:in `each'
D:/Applications/developpement/Ruby/lib/ruby/gems/1.9.1/gems/compass-0.11.beta.2/lib/compass/sass_extensions/functions/sprites.rb:146:in `construct_sprite'
D:/Applications/developpement/Ruby/lib/ruby/gems/1.9.1/gems/compass-0.11.beta.2/lib/compass/sass_extensions/functions/sprites.rb:126:in `generate'
D:/Applications/developpement/Ruby/lib/ruby/gems/1.9.1/gems/compass-0.11.beta.2/lib/compass/sass_extensions/functions/sprites.rb:294:in `sprite_url'
D:/Applications/developpement/Ruby/lib/ruby/gems/1.9.1/gems/compass-0.11.beta.2/lib/compass/sass_extensions/functions/sprites.rb:220:in `method_missing'
D:/Applications/developpement/Ruby/lib/ruby/gems/1.9.1/gems/compass-0.11.beta.2/lib/compass/sass_extensions/functions/sprites.rb:211:in `to_s'
D:/Applications/developpement/Ruby/lib/ruby/gems/1.9.1/gems/sass-3.1.0.alpha.221/lib/sass/script/list.rb:36:in `block in to_s'
D:/Applications/developpement/Ruby/lib/ruby/gems/1.9.1/gems/sass-3.1.0.alpha.221/lib/sass/script/list.rb:36:in `map'
[...]

PR #633, opened 2011-11-28
fixing total sprite width for repeat-x images

vbichkovsky:
It is related to the issue #296. I calculate lowest common multiple for all repeat-x images in the sprite and then correct calculated @width, so that it divides by lcm without reminder.

### DIFF
--- a/lib/compass/sass_extensions/sprites/sprite_methods.rb
+++ b/lib/compass/sass_extensions/sprites/sprite_methods.rb
@@ -1,3 +1,5 @@
+require 'rational'
+
 module Compass
   module SassExtensions
     module Sprites
@@ -27,11 +29,15 @@ module Compass
         
         # Creates the Sprite::Image objects for each image and calculates the width
         def init_images
+          repeated_widths = []
           @images = image_names.collect do |relative_file|
             image = Compass::SassExtensions::Sprites::Image.new(self, relative_file, kwargs)
+            repeated_widths << image.width if image.repeat == 'repeat-x'
             @width = [ @width, image.width + image.offset ].max
             image
           end
+          lcm = repeated_widths.inject {|l, w| l.lcm(w)}
+          @width += @width % lcm if lcm
         end
         
         # Calculates the overal image dimensions


scottdavis:
oh nevermind it helps if i read the ticket before the source fail

vbichkovsky:
no problem, I can comment what I did easily :)
1. require 'rational' - this is necessary for Ruby 1.8.7, to get access to the 'lcm' method.
2. I iterate through all repeat-x images in the sprite map and calculate lowest common multiple (lcm) of their widths. For example, if their widths are [1, 2, 3, 4] then lcm = 12.
3. I adjust sprite width to be a factor of lcm (if max width = 13, then I make it 24).

Btw, it can be written better than in my code.
m = 1
(inside image_names.collect loop)
  if image.repeat == 'repeat-x' then m = m.lcm(image.width)
(end of loop)
@width += @width % m

scottdavis:
ok trying to bang this out tonight

scottdavis:
i dono if thats entirely correct since the width is the widest image so say:
goal here is a with of 6

image_widths =  [1,2,3,4,5] 
@width = image_widths.max
image_three = 'repeat-x'
so shouldn't the width in this case be 6?

since bumping the image by 1 pixel will allow an image of width 3 to repeat-x to the edge?

using sudo the code above i get 7

scottdavis:
and using 4 as my width and 3 as the image im trying to get i get 5 which would clip the repeating image

scottdavis:
<pre>
        def modify_image_width_for_repeat_x!
          m = @images.inject(1) do |m, img| 
            if img.repeat_x?
              m.lcm(img.width) 
            else 
              m + 0
            end
          end
          @width += @width % m
        end
</pre>

scottdavis:
take the smallest multiple of the LCM of the repeating ones that's at least equal to the non-repeating ones' max is the solution here i just am having a brain fart and can't implement it correctly lol

vbichkovsky:
oops, sorry :) 
I wonder why it worked in my case. Here is correction:

m = images@.inject(1) {|m,i| img.repeat_x? ? m.lcm(img.width) : m}
remainder = @width % m
@width += (m - remainder) unless remainder.zero?

scottdavis:
well that was simple =) i hate math some days

PR #651, opened 2011-12-19
if manifest_file not exists, discover :all

tka:


tka:
need test case and update docs

scottdavis:
also would you mind redoing this against master?

tka:
@scottdavis I think the master branch can use the same patch, should I send another pull request for master branch?

scottdavis:
Stable is always the current stable. I don't think we are planing a bug fix to the 11.x and are moving forward with 0.12.x so yes please do this against master

tka:
@scottdavis ok, I understand now. thank you

PR #652, opened 2011-12-19
if extension missing templates/project/manifest.rb, use discover :all

tka:


features/step_definitions/extension_steps.rb
```diff
@@ -18,6 +18,13 @@
   end
 end
 
+Given /^I have a fake extension without manifest file at (.*)$/ do |directory|
+  directory.gsub!('~', ENV["HOME"]) if directory.include?('~/')
```
scottdavis, line 22: I donno if i like this in a test case using my home directory to run tests in is kinda making me queezy we probaly should mock this in tests /cc @chriseppstein

chriseppstein, line 22: ya that's probably right.

scottdavis:
going t cherry-pick these changes in later im scared to pull this in with tests accessing the home directory without mocking

chriseppstein:
I merged this in without the test.

PR #657, opened 2011-12-28
Improves index/functions - adds links to sass functions and differentiat...

MarioRicalde:
...e them from Compass ones.

Not super sexy implementation of the Sass Functions, but it'll do a trick till we find a better solution.


scottdavis:
what exactly is this doing?

Issue #634, opened 2011-11-30
Rails 3.1 and sprite-dimensions


andriusch:
Seems on newest alpha of compass (0.12.alpha.2) sprite-dimensions doesn't work anymore (it works on 0.12.alpha.0). Error looks like this:

TypeError: can't convert nil into String
/home/box2/.rvm/gems/ruby-1.9.3-p0/gems/compass-0.12.alpha.2/lib/compass/sass_extensions/functions/image_size.rb:17:in `extname'

I investigated this and seems that wrong image path is passed to Rails.application.assets.find_asset (eg. compass passes app/assets/image/icons/checked.png, while it should pass icons/checked.png.

On a bit unrelated note there seems to be a typo in image_dEmensions method name.

scottdavis:
are you using sprockets?

scottdavis:
Also the image_dimensions needs to get the full path to the file. im not sure why it would be passing anything to sprockets. It needs to ask sprockets where to find the file? maybe if i could see a gist of the sass causing the issue i can understand this a bit better.

ghost:
Yes using sprockets. I don't have code right now, will post it tomorrow if I won't forget, but it's nothing fancy just creating sprite map with sprite-map and then asking for dimensions of one of the sprites using image-dimensions.

If you take a look at https://github.com/chriseppstein/compass/blob/master/lib/compass/app_integration/rails/actionpack31/helpers.rb#L13 this is where compass seems to ask Rails (sprockets) for full path to image, but this fails as compass passes path to image from rails root, while it should pass it with app/assets/images removed.

scottdavis:
weird im not having this issue in my app using the new version and im using tons of sprites is it possibly an old version of sprockets?

scottdavis:
https://github.com/chriseppstein/compass/blob/master/lib/compass/app_integration/rails/actionpack31/helpers.rb#L13

This function needs the full system path to the image so it can get the dimensions im not sure what you are saying? any update on this?

andriusch:
I mean this: Rails.application.assets.find_asset(image_file), it needs path without app/assets/images. I'm using:
sprockets (2.0.3)
rails (3.1.3)
compass (0.12.alpha.2)

Here's code:

``` scss
$cbt-v-space: 20;
$cbt-h-space: 30;
$cbt-icons: sprite-map("cbt-icons/*.png",
    $cleanup: true,
    $checked-position: $cbt-h-space * 4, $checked-spacing: $cbt-v-space,
[...]

PR #662, opened 2012-01-02
Correctly detect if the assets pipeline is enabled.

ShogunPanda:
I've got the same error of issue #634.
In my case, the problem was that compass included `ConfigurationDefaultsWithAssetPipeline` even if in `config/application.rb` I set 

``` ruby
config.assets.enabled = false
```

I modified the condition by adding an additional method which detect if the assets pipeline is enabled.

Hope this helps!
   Shogun

### DIFF
--- a/lib/compass/app_integration/rails.rb
+++ b/lib/compass/app_integration/rails.rb
@@ -26,7 +26,7 @@ module Compass
       def configuration
         config = Compass::Configuration::Data.new('rails')
         config.extend(ConfigurationDefaults)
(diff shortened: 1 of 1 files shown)


scottdavis:
Im currently moving this out of compass into compass-rails im not going to merge this here but i will put it in there

scottdavis:
https://github.com/Compass/compass-rails/commit/4f0d5f00fdc4f23c3ea710e1f46e3e67ad36b146

PR #666, opened 2012-01-08
Fix wrong degree to radian conversion in trig function. Might break some...

ShishKabab:
When passing degrees to a trig function, the angle is halved because of a wrong denominator (360 instead of 180) while converting from degrees to radians. See http://www.teacherschoice.com.au/maths_library/angles/angles.htm . Some have worked around this by just doubling their angles, so this change might break that workaround.


chriseppstein:
I can't believe I effed this up. grrr.

PR #671, opened 2012-01-11
Added reset for button tag in order to fix the FF bug

djadriano:
Fixed the button's extra padding caused by an obscure Firefox bug found by other guys too(http://stackoverflow.com/questions/4607410/sliding-door-button-element-broke-in-firefox-only).

### DIFF
--- a/frameworks/compass/stylesheets/compass/reset/_utilities.scss
+++ b/frameworks/compass/stylesheets/compass/reset/_utilities.scss
@@ -11,8 +11,8 @@
   dl, dt, dd, ol, ul, li,
   fieldset, form, label, legend,
   table, caption, tbody, tfoot, thead, tr, th, td,
-  article, aside, canvas, details, embed, 
-  figure, figcaption, footer, header, hgroup, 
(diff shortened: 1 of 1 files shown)


scottdavis:
ping @chriseppstein @ericam @imathis any inout on this?

mirisuzanne:
I like having the mixin available, and even a simple way to apply it upfront to `button, button::-moz-focus-inner`, but I'm not sure it's a good idea to apply it by default (see best-practice on `:focus` styles). It's best to have some default button styles unless you override them explicitly.

Also, if we're resetting buttons, this doesn't go the whole way. I still have background, font and color settings from the browser. The reset I use includes `background: none; color: inherit; font: inherit;` to complete the job.

scottdavis:
going to agree with eric the mixin is fine if improved but don't include it in the default reset.

chriseppstein:
@djadriano what's your plan here? Do you want to iterate on this patch?

chriseppstein:
Closing for now.

Issue #619, opened 2011-11-16
Where does the output of `compass watch` go?


timmolendijk:
I'm trying to run `compass watch` from a `Cakefile` (CoffeeScript's Node.js-based build system). When I spawn a new process (`require('child_process').spawn 'compass', ['watch']`) it will indeed watch changes and recompile them. In that sense it works. But I cannot seem to catch the process's output. Nothing comes in over `stdout`.

This problem only applies to `compass watch`, output of other commands such as `compass compile` passes `stdout` as expected.

What is so special about `compile watch` that it behaves differently, and how can I get hold of the process's output?

Why doesn't its output go through `stdout` as most people would expect?

chriseppstein:
It outputs to stdout like all the rest of the commands. If it didn't you wouldn't see it in the terminal output. 

Hunt & pecked on my iPhone... Sorry if it's brief!

timmolendijk:
Yeah, that's what I was thinking. Node.js's spawn method doesn't seem to agree though.

It turns out to be even weirder. With `--boring` nothing comes in over stdout or stderr for sure. Without `--boring` some data seems to come in but it is incapable of making something of it. Whatever you do it prints as empty string (but some coloring seems to make it through).

_Any_ clue what's different in `watch` compared to `compile` that might be causing this difference in behavior?? Like I said when running `compile` all works perfect -- with and without `--boring`, including colors.

Also, what strikes me as odd is that any output starting with `>>>` (for example `>>> Compass is polling for changes. Press Ctrl-C to Stop.`) is nowhere to be found in either scenario. Not even with `compile`, in which case all the other output (like `unchanged sass/style.scss`) _does_ make it through. There must be a difference in how these strings are being evoked from `compass`?!?

timmolendijk:
P.S. The same behavior exists when spawning `compass` in a plain Node.js script, so Cake(file)/CoffeeScript has nothing to do with this.

timmolendijk:
More analysis:
- forked processes shouldn't be problematic: https://twitter.com/#!/ejpbruel/status/136636397246353408
- Node's `child_process.exec` experiences similar problems: https://twitter.com/#!/ejpbruel/status/136639257161900032

Really _really_ curious what `compass` is doing internally... Any thoughts? My Ruby skills are non-existent unfortunately.

timmolendijk:
[Issue report over at Node.js](https://github.com/joyent/node/issues/2127)

chriseppstein:
[...]

PR #675, opened 2012-01-16
Flush stdout buffer to make sure the stdout message show up correctly.

jezdez:
Not sure if that's the correct way of handling the problem, but I've certainly could reproduce it (by having a supervisor conf with a `compass watch .` program configured).

Issue #619 has probably a better explanation.


scottdavis:
Compass probably needs to implement a logging API across the entire
Project these keep popping up now and then I'll add it to my list of things to look into

jezdez:
That would certainly also solve the problem, true.

superstructor:
Its been a year, #691 is not done and forking compass as described in #619 is still broken. Can you please merge this change ? It may not be the ideal comprehensive solution but at least its not completely broken for those who need to fork compass from a build process and consume its output. Thanks :-)

nick-jonas:
Agreed, could really use this!

PR #684, opened 2012-01-24
Update lib/compass/app_integration/rails.rb

bmaltzan:
typo: intializer -> initializer


scottdavis:
This has been removed in in 0.12 but thanks for finding it!

PR #708, opened 2012-02-08
Modified "background" property to be "background-image"

Xananax:
I had sprites with transparency that I used throughout the website, using different background-colors in different places, specially with :hover and :active selectors. The current implementation of the sprite helper used "background: $sprite no-repeat" which was overriding my background-color property; I changed it to:
background-image:$sprite, background-repeat:no-repeat, which allows more granular setting-up of the background property.

### DIFF
--- a/lib/compass/sprite_importer.rb
+++ b/lib/compass/sprite_importer.rb
@@ -140,7 +140,8 @@ $#{name}-clean-up: true !default;
 // All sprites should extend this class
 // The #{name}-sprite mixin will do so for you.
 \#{$#{name}-sprite-base-class} {
-  background: $#{name}-sprites no-repeat;
+  background-image: $#{name}-sprites;
+  background-repeat: no-repeat;
 }
 
 // Use this to set the dimensions of an element


chriseppstein:
seems ok. Can you please update the tests? http://compass-style.org/help/tutorials/contributing/

scottdavis:
This will not apply to master cleanly since this has been rewritten if you could make the change here (https://github.com/chriseppstein/compass/blob/master/lib/compass/sprite_importer/content.erb#L45) and re submit a pull request against master please!

Xananax:
Sorry, I use git & hg for my personal use but I am not used to work in
group environments. I have only a vague idea of how this is supposed to
work. Also, I never used Ruby, save for compiling compass. After hours of
dependencies problems, I finally got _"An error occured while installing
rb-appscript (0.6.1), and Bundler cannot continue."_
For these reasons, I fear I might do more harm than good. I know my patch
is insignificant and does not deserve more than 5mn attention,
but this is my attempt to ease your lives in return for the great work
you're giving to the community; if it reveals more cumbersome than helpful
because of my inexperience, I'll just remove everything and submit a
feature request (unless you think it is important that I continue).
I could not run the test because of the error stated above, so I can only
surmise that they work.
Sorry if I am a bother.

PR #712, opened 2012-02-09
Sprite background image

Xananax:
Changed the "background" property to "background-image" and "background-repeat", in order to allow one to have more granular control over the background property (set both a background image and, say, a background-color).

### DIFF
--- a/lib/compass/sprite_importer/content.erb
+++ b/lib/compass/sprite_importer/content.erb
@@ -42,7 +42,8 @@ $<%= name %>-inline: false !default;
   } 
 } @else {
   #{$<%= name %>-sprite-base-class} {
-    background: $<%= name %>-sprites no-repeat;
+    background-image: $<%= name %>-sprites;
+    background-repeat: no-repeat;
   }
 }
 //sass functions to return the dimensions of a sprite image as units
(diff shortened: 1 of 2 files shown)


chriseppstein:
Thank you for persisting!

Xananax:
It's nothing!
Hope my changes are at least consistent with coding style/expectancy and won't add another task for you. I tried to follow the indications of the "contribute" page as closely as possible, but all this is new to me.
I hope to be able to contribute more. I just discovered compass and the compiled time it has saved me must amount to a week at least.
Please do not hesitate to tell me if I did anything wrong.

scottdavis:
actually i had to revert this all the tests were breaking see: http://travis-ci.org/#!/chriseppstein/compass/jobs/653171

Xananax:
As I said I never used Ruby, I have no idea how rake/etc works. I kind of understand the code because, well, it's code, but I am not sure of what I am doing at all.
I think it would be best if someone else took care of the tests; if there is no one, I will get back to it, but I have a project that is nearing deadline and I can't give this more time for the moment. I will try to read some ruby basics during this week, and see how your test system works.
Sorry I can't be more helpful

PR #715, opened 2012-02-13
fix transparent gradients

curzonj:
This is by no means the best way to solve the problem, just the only
that I found on google.

https://gist.github.com/1722819

I need this so that I can include the twitter bootstrap files into my application.css file.


scottdavis:
The more i look at this the more im thinking this is the wrong approach sass really needs a "transparent" color

mwawrusch:
+1000000 - just wasted a couple hours on this and similar issues.

chriseppstein:
I have a patch for this that preserves the transparent keyword in the output.

chriseppstein:
But thank you very much for your patch!

Issue #289, opened 2011-03-01
Bug with transition mixin + transform as property


aurelien-baudet:
Hi!

It's not possible to do this:
@include transition-property(transform)

In order to have:
-webkit-transition-property: -webkit-transform;
-moz-transition-property: -moz-transform
...

Instead it generates:
-webkit-transition-property: transform;
-moz-transition-property: transform;
...

Of course the generated css doesn't work

Thanks

chriseppstein:
hmmm. That's certainly annoying. I'd rather not bake-in a list of what properties are prefixed and what aren't. What if there was a `transition-prefixed-property` mixin?

aurelien-baudet:
Yes you can use a special mixin for that, the important is to have the good result for each browser

doctyper:
I was actually just working on a patch for this when I came across this issue. Patch:
https://github.com/doctyper/compass/commit/50c7c551408f3a3e10daa2ab33d55e4a11c3e8c3

As you can see it takes a lot of customization (unless I'm missing some helper functions?) Unfortunately I did have to "bake-in" prefixed properties in order to get the patch to work.

@chriseppstein Thoughts? Would this be a valid workaround or would your suggestion make more sense?

ZeeAgency:
So... what solution do we have today ?

davidcalhoun:
+1

Hope this is fixed soon.

scottdavis:
ping @paulirish you have any input on the way to best implement this? hate to bother you but you seem to have a greater grasp on vendor prefix's then i do

paulirish:
Looking at the hardcoded list of props that are both 1) animatable/transitionable and 2) prefixed we don't end up with many that people are using, aside from `transform`

`background-size` is unprefixed everywhere except ff3.6, same with `box-shadow`. and `font-size-adjust` is ff only and probably not animatable..

(yup. it's not: http://oli.jp/2010/css-animatable-properties/  )

so tbh I think we should just special case "transform" when used in a transition and apply the prefixes automatically.

wagenet:
@paulirish Sounds good. Would love to see this implemented just for the transform case.

paulirish:
Can someone on this thread take a stab at a pull req for this?

chriseppstein:
Oh look a whole guide for contributing to Compass! http://compass-style.org/help/tutorials/contributing/

paulirish:
Fixed now as #728 was pulled?

`\o/`

davatron5000:
[...]

PR #725, opened 2012-02-17
Add support for transitionable prefixed properties

doctyper:
Addresses issue #289.

The transition mixin now adds appropriate prefixes for both transform and transform-origin properties. The mixin is also built to handle any additional prefixed properties the W3C throws at us down the line.

@chriseppsten I built a couple of helper functions for array / string manipulation. Let me know if I'm duplicating any already-existing functionality.


doctyper:
Hm, well that was unexpected. Anyone know how to decouple pull requests?

chriseppstein:
You'll need to make a separate branch for each feature. See the git `cherry-pick` command.

PR #728, opened 2012-02-18
Adding support for prefixed transition properties

doctyper:
Addresses issue #289.

The transition mixin now adds appropriate prefixes for both transform and transform-origin properties. The mixin is also built to handle any additional prefixed properties the W3C throws at us down the line.

@chriseppstein I built a couple of helper functions for array / string manipulation. Let me know if I'm duplicating any already-existing functionality.

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_transition.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_transition.scss
@@ -32,7 +32,7 @@ $transitionable-prefixed-properties: transform, transform-origin !default;
 
 @mixin transition-property($properties: $default-transition-property) {
   @if type-of($properties) == string { $properties: unquote($properties); }
-  $raw-properties: -compass-space-list($properties);
+  $raw-properties: ($properties);
 
   $webkit-transition-property: false;
   $moz-transition-property: false;
@@ -181,7 +181,7 @@ $transitionable-prefixed-properties: transform, transform-origin !default;
   $transition-10: false
 ) {
   @if $transition-1 == default {
-    $transition-1 : -compass-space-list(compact($default-transition-property, $default-transition-duration, $default-transition-function, $default-transition-delay));
+    $transition-1 : (compact($default-transition-property, $default-transition-duration, $default-transition-function, $default-transition-delay));
   }
 
   $transition : compact($transition-1, $transition-2, $transition-3, $transition-4, $transition-5, $transition-6, $transition-7, $transition-8, $transition-9, $transition-10);
@@ -207,7 +207,7 @@ $transitionable-prefixed-properties: transform, transform-origin !default;
 }
 
 @mixin prefix-friendly-transition($transition) {
-  $raw-transition: -compass-space-list($transition);
+  $raw-transition: ($transition);
 
   $webkit-transition: false;
   $moz-transition: false;
--- a/frameworks/compass/stylesheets/compass/css3/_transition.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_transition.scss
@@ -23,15 +23,15 @@ $default-transition-function: false !default;
 
 $default-transition-delay: false !default;
 
+$transitionable-prefixed-properties: transform, transform-origin !default;
+
 // One or more properties to transition
 //
 // * for multiple, use a comma-delimited list
 // * also accepts "all" or "none"
 
 @mixin transition-property($properties: $default-transition-property) {
-  @include experimental(transition-property, unquote($properties),
-    -moz, -webkit, -o, -ms, not -khtml, official
-  );
+  @include prefix-friendly-transition(unquote($properties));
 }
 
 // One or more durations in seconds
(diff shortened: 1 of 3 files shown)


paulirish:
Yessssss

Richard i love you

I think you should add in the official unprefixed version here. 

 also it's not place to ask but I imagine it'd be nice to have a test written for this feature.

doctyper:
@paulirish The unprefixed version is covered using the [@experimental mixin](https://github.com/chriseppstein/compass/pull/728/files#L0L130). I just landed test cases for the changes as well. Good thing too, since some of my previous changes were apparently broken.

Also, not sure if you guys agree but my changes are now on the beefy side. @chriseppstein is there anything I can do to slim this down?

frameworks/compass/stylesheets/compass/css3/_transition.scss
```diff
@@ -78,29 +100,56 @@ $default-transition-delay: false !default;
   $function: $default-transition-function,
   $delay: $default-transition-delay
 ) {
+  $match: false;
+
+  @each $prefixed-property in $transitionable-prefixed-properties {
+    @if $property == $prefixed-property {
+      $match: true;
+    }
+  }
```
chriseppstein: Setting `$match` should be doable using functions instead of loops. But it's not -- we need better list and string functions. >_<

frameworks/compass/stylesheets/compass/css3/_transition.scss
```diff
@@ -23,15 +23,37 @@ $default-transition-function: false !default;
...
+    @if $properties == $prefixed-property {
+      $match: true;
+    }
+  }
+
+  @if $match {
+    @if $experimental-support-for-webkit    { -webkit-transition-property: -webkit-#{$properties}; }
+    @if $experimental-support-for-mozilla   { -moz-transition-property: -moz-#{$properties}; }
+    @if $experimental-support-for-microsoft { -ms-transition-property: -ms-#{$properties}; }
+    @if $experimental-support-for-opera     { -o-transition-property: -o-#{$properties}; }
```
chriseppstein: pretty sure this fails if $properties is a list. We need to map the prefix onto each of the values of the list.

frameworks/compass/stylesheets/compass/css3/_transition.scss
```diff
@@ -23,15 +23,37 @@ $default-transition-function: false !default;
...
+    @include experimental(transition-property, $properties,
+      not -moz, not -webkit, not -o, not -ms, not -khtml, official
+    );
```
chriseppstein: There's really no point in calling the `experimental` mixin here if you're just setting the official version.

frameworks/compass/stylesheets/compass/css3/_transition.scss
```diff
@@ -23,15 +23,37 @@ $default-transition-function: false !default;
...
-  );
+  @if type-of($properties) == string { $properties: unquote($properties); }
+
+  $match: false;
+
+  @each $prefixed-property in $transitionable-prefixed-properties {
+    @if $properties == $prefixed-property {
+      $match: true;
+    }
+  }
```
chriseppstein: Setting $match should be doable using functions instead of loops. But it's not -- we need better list and string functions. >_<

frameworks/compass/stylesheets/compass/css3/_transition.scss
```diff
@@ -78,29 +100,56 @@ $default-transition-delay: false !default;
...
-        official
-      );
+      @if $match {
+        @if $experimental-support-for-webkit    { -webkit-transition: -webkit-#{$property} $duration $function; -webkit-transition-delay: $delay; }
+        @if $experimental-support-for-mozilla   { -moz-transition: -moz-#{$property} $duration $function $delay; }
+        @if $experimental-support-for-microsoft { -ms-transition: -ms-#{$property} $duration $function $delay; }
+        @if $experimental-support-for-opera     { -o-transition: -o-#{$property} $duration $function $delay; }
+        @include experimental(transition, $property $duration $function $delay,
+          not -moz, not -webkit, not -o, not -ms, not -khtml, official
+        );
```
chriseppstein: Again, no point in calling `experimental`

frameworks/compass/stylesheets/compass/css3/_transition.scss
```diff
@@ -78,29 +100,56 @@ $default-transition-delay: false !default;
...
-        official
-      );
+      @if $match {
+        @if $experimental-support-for-webkit    { -webkit-transition: -webkit-#{$property} $duration $function; }
+        @if $experimental-support-for-mozilla   { -moz-transition: -moz-#{$property} $duration $function; }
+        @if $experimental-support-for-microsoft { -ms-transition: -ms-#{$property} $duration $function; }
+        @if $experimental-support-for-opera     { -o-transition: -o-#{$property} $duration $function; }
+        @include experimental(transition, $property $duration $function,
+          not -moz, not -webkit, not -o, not -ms, not -khtml, official
+        );
```
chriseppstein: Again, no point in calling experimental

frameworks/compass/stylesheets/compass/css3/_transition.scss
```diff
@@ -126,8 +175,84 @@ $default-transition-delay: false !default;
...
+  @if $existing {
+    @if $addition {
+      $existing: $existing $addition;
+    }
+  } @else {
+    $existing: $addition;
+  }
+
+  @return $existing;
+}
```
chriseppstein: This actually returns a list -- not a string. Seems like we can use `compact(join($existing, $addition))` which are already built into compass and sass.

frameworks/compass/stylesheets/compass/css3/_transition.scss
```diff
@@ -126,8 +175,84 @@ $default-transition-delay: false !default;
...
+  } @else {
+    @if $array {
+      $array: $array $object;
+    } @else {
+      $array: $object;
+    }
+  }
+
+  @return $array;
+}
```
chriseppstein: Seems the sass `append` function should suffice. Am I missing something?

chriseppstein: - `append($list, $value, comma)`
- `append($list, $value, space)`
- `append($list, $value, auto)`

frameworks/compass/stylesheets/compass/css3/_transition.scss
```diff
@@ -126,8 +175,84 @@ $default-transition-delay: false !default;
...
+    }
+  } @else {
+    $existing: $addition;
+  }
+
+  @return $existing;
+}
+
+@mixin prefix-friendly-transition($transition) {
+  $raw-transition: -compass-space-list($transition);
```
chriseppstein: we don't need to use `-compass-space-list` anymore. Sass provides `()` to construct an empty space-delimited list.

chriseppstein: Oh. sorry. This coerces a possible string into a list. Seems like the string case shouldn't exist anymore. But I'm foggy on the deprecation status.

frameworks/compass/stylesheets/compass/css3/_transition.scss
```diff
@@ -126,8 +175,84 @@ $default-transition-delay: false !default;
...
+
+  $webkit-transition: false;
+  $moz-transition: false;
+  $ms-transition: false;
+  $o-transition: false;
+
+  @each $single-transition in $raw-transition {
+    $property: nth($single-transition, 1);
```
chriseppstein: I don't understand this. are there cases where we have a list of lists?

chriseppstein: And if so, why, and is it correct to discard all but the first element?

frameworks/compass/stylesheets/compass/css3/_transition.scss
```diff
@@ -126,8 +175,84 @@ $default-transition-delay: false !default;
...
+    $match: false;
+
+    @each $prefixed-property in $transitionable-prefixed-properties {
+      @if $property == $prefixed-property {
+        $match: true;
+      }
+    }
```
chriseppstein, line 123: 3 times! MUST BE REFACTORED

frameworks/compass/stylesheets/compass/css3/_transition.scss
doctyper: The transition property allows for multiple transition declarations:

```
transition: transform 3s ease-out, opacity 2s ease-in;
```

Here I'm iterating through each transition and manipulating the first value in the list (transform, opacity, etc).

frameworks/compass/stylesheets/compass/css3/_transition.scss
doctyper: The issue here is that I need to append values to an initially-false variable:

```
$list: false;
append($list, foo, comma);
```

This results in:

```
$list: false, foo;
```

I expect:

```
$list: foo;
```

The only workaround I've come up with is to set the initial $list value to an unquoted string. Unfortunately, this adds an empty space to the beginning of the value and fails during testing:

```
$list: unquote("");
append($list, foo, comma);
```

Result:

```
$list:  foo;
```

Expected (sans space):

```
$list: foo;
```

Any ideas?

chriseppstein: The compass `compact` function will remove false values from a list.

frameworks/compass/stylesheets/compass/css3/_transition.scss
chriseppstein: gotcha.

frameworks/compass/stylesheets/compass/css3/_transition.scss
doctyper, line 123: Agreed, and done.

doctyper:
Just pushed an update to address most/all of your concerns @chriseppstein. Important bits:
- transform-property now takes multiple properties
- removed redundant helper mixins
- refactored code to DRY things up

chriseppstein:
Thanks will review again soon.

PR #733, opened 2012-02-25
Add compass and blueprint to Sass load_path

jch:
Following from the discussion of on [this sprockets ticket](https://github.com/sstephenson/sprockets/issues/283), Compass does not add itself to Sass's load path by default.

I'm trying to use Compass standalone and `@import compass` without specifying a config file and having Sprockets and Sass do the compilation. Since compass isn't on the load path, I have to manually add it. Seeing how compass depends on sass, I thought it'd be simpler if compass appends itself to sass's load path by default.  I've attached a sample code snippet that does this.

Do you think this is a good idea? If so, I can add a test for the attached code.

### DIFF
--- a/lib/compass.rb
+++ b/lib/compass.rb
@@ -33,3 +33,9 @@ end
 
 # for rails upgrade warnings in 0.12
 require 'compass/app_integration/rails'
+
+# append compass stylesheets to Sass load path by default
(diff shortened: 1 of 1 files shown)


chriseppstein:
This should use the new public API that Sass added for this reason once Sass 3.2 is out. https://github.com/nex3/sass/commit/5c93cbfe95c8ede75ede327cd0892a32c1207a1b

PR #734, opened 2012-02-25
Couple of fixes for tests

Flameeyes:
As the summary says.


scottdavis:
The tests on stable are currently broken. is it intended that these are failing tests? if so can you move them to a pull request against master?

Flameeyes:
I have one more failure (actually a stack overflow), but with this I get no test failure.

chriseppstein:
Thanks, all these changes are now on stable.

Issue #738, opened 2012-02-27
Validate sprite css in mixins not helpers


zanker:
Currently, Compass will validate the CSS of a sprite as soon as the image is imported.  The problem with this is that it makes it impossible to use images with numbers at the start because it's an invalid CSS tag.

This only causes invalid CSS  if you use the mixins referenced on the [spriting tutorial](http://compass-style.org/help/tutorials/spriting/), but if you use the [advanced helpers](http://compass-style.org/reference/compass/helpers/sprites/) then there are no actual validation errors.

Unless this is intentional, I would be happy to write a pull request to accomplish this, but I was unable to find the exact portion that defines all of the CSS generating helpers in the tutorial.

scottdavis:
This is intentional.

chriseppstein:
Note, you can use sprites without using the "magic" import. I did this recently for my app:

``` sass

@import "compass/utilities/sprites/base"

$step-sprites: sprite-map("local_wizard/step/*.png", $layout: vertical, $cleanup: true)
.step-sprite
  background: white $step-sprites no-repeat

body.dumbledore
  #app-collection-wrapper.who
    @extend .step-sprite
    +sprite($step-sprites, lilacs)
  #app-collection-wrapper.what
    @extend .step-sprite
    +sprite($step-sprites, heart)
  #app-collection-wrapper.where
    @extend .step-sprite
    +sprite($step-sprites, flowerbox)
  #app-collection-wrapper.when
    @extend .step-sprite
    +sprite($step-sprites, stairs)
  #app-collection-wrapper.payment
    @extend .step-sprite
    +sprite($step-sprites, orangeflower)
  #app-collection-wrapper.contact-info
    @extend .step-sprite
    +sprite($step-sprites, letter)
```

If you take this approach, the naming requirements are not necessary to enforce strictly. Changing the way this error is raised seems reasonable.

zanker:
Sorry the original description should have included an example for clarity. How I was calling it is identical to your example.

Was able to track down the files I needed, I'll see about getting a pull request up tomorrow.

chriseppstein:
@zanker sounds good.

zanker:
https://github.com/chriseppstein/compass/pull/739

PR #739, opened 2012-02-27
Allow numeric filenames in sprites

zanker:
From #738, this will allow numeric filenames in Compass by checking if the full selector is valid CSS rather than just the tag name, so "16x16/100.png" would still be an invalid sprite because it turns into ".16x16-100", but "icons/100.png" would be fine because it turns into ".icons-100".

For the selector validation in the scss file, I'm not entirely a fan of it being there as a new method rather than using the existing ones. But it seemed like the best solution to avoid hacking it into the existing sprite_has_selector and sprite_does_not_have_parent methods.


chriseppstein:
As an aside: I think Sass needs a way to cause an error from inside a sass file, so that the `sprite_has_valid_selector` function can just return a boolean.

This patch looks good to me. @scottdavis do you +1?

PR #743, opened 2012-02-28
Refactor hide-text

scottkellum:
This fixes two major issues with the -9999px indent. With this method
the text is always flowing away from the container so there is no risk
of it flowing into view. Also it prevents a 9999px box from being
rendered by the browser saving tons of video memory and making things
faster ( see http://youtu.be/xuMWhto62Eo )

### DIFF
--- a/doc-src/content/stylesheets/partials/_home.scss
+++ b/doc-src/content/stylesheets/partials/_home.scss
@@ -4,7 +4,8 @@ body#home {
   h1#logo {
     $logo: 'compass-logo.png';
     background: image-url($logo) no-repeat;
-    text-indent: -9999px;
+    text-indent: 100%;
+    white-space: nowrap;
     overflow: hidden;
     width: image-width($logo);
     height: image-height($logo)/2;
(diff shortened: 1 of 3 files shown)


scottkellum:
Possible issues:
- @chriseppstein pointed out this may have issues with inline elements (https://twitter.com/chriseppstein/status/174505762213736448)
- Some fonts may have negative metrics causing them to edge into view. Increasing the indent may be needed.

scottkellum:
making the indent 150% should solve most font metric issues enforcing at least a 1px margin of error on a 1px wide element.

Inline elements don’t work with either method. Inline-block elements exhibit different behavior when a width is not applied to them. The -9999px method causes the element to collapse while the 150% method causes the element to grow to include the content. This is an edge case, not sure if it is worth worrying about.

kaelig:
Love it, very elegant and performant, thanks!

scottkellum:
Also follow this discussion: https://github.com/h5bp/html5-boilerplate/issues/1005

paulirish:
:+1: yay team

antsa:
I encoutered a problem with this new image replacement technique and IE7. An isolated test case and explanation can be found here: http://jsfiddle.net/salantsa/3DYd2/

PR #749, opened 2012-03-01
Update with Kellum Method _replacement.scss

stephenway:
Replace the -9999px Hack with the Kellum Method http://bit.ly/Aazp14
• Really long strings of text will never flow into the container because they always flow away from the container.
• Performance is dramatically improved because a 9999px box is not drawn. Noticeably so in animations on the iPad 1.


stephenway:
Just saw this https://github.com/chriseppstein/compass/pull/743
Closing now.

PR #750, opened 2012-03-01
How Compass could use Scott Kellum's updated Image Replacement

nearestnabors:
See Zeldman: http://www.zeldman.com/2012/03/01/replacing-the-9999px-hack-new-image-replacement/

The new method improves browser performance.

Made hide-text an @extend rather than a @mixin to help keep the compiled CSS lighter. You might want to keep it as a mix-in, but what's a mixin without variables if not a bulky @extend?


chriseppstein:
Thanks for the pull request! It's a dupe of #743. Also I like what you're doing with the use of `@extend`, but I want to wait to embrace base classes until Sass 3.2's placeholder selectors are available to us.

Issue #752, opened 2012-03-05
Fonts .eot & .svg are not "discovered"


MoOx:
I'm creating a compass extension & have & problem when I try to use discover.
My fonts are not copied as they should be.
.eot (for IE) & .svg (iPhone i think) are not copied as fonts. I know .svg is complex to manager, but .eot should be quick to add right ? I've fork but I don't find where to add this in the code... :s

PR #753, opened 2012-03-06
Add .eot (Embedded OpenType) to fonts format

MoOx:
Tiny PR.

Fix part of #752

I did not add .svg since it's not a real font format & it can create issues.


MoOx:
Btw maybe I could add the mime type here ?
https://github.com/chriseppstein/compass/blob/stable/lib/compass/sass_extensions/functions/inline_image.rb 

note : .eot mime type "application/vnd.ms-fontobject"

PR #754, opened 2012-03-07
Correct Mime type for woff font files

aaronchi:
This avoids a warning in Chrome when using embedded fonts


aaronchi:
sorry. this should be commited to master :(

PR #758, opened 2012-03-08
Add more math functions and unite them with Trig

silvenon:
The first commit adds more math functions and groups them with trigonometry functions as `Math` module, i.e. `Trig` is merged into `Math`.

The second commit adds those span elements around some arguments in helpers which are missing them. I know it's not really related to math functions, but I wanted to squeeze those in also :)


chriseppstein:
Can you add tests for the new functions like we have for the trig functions: https://github.com/chriseppstein/compass/blob/stable/test/units/sass_extensions_test.rb#L67-82

silvenon:
Unfortunately, I can't run them right now because of some gem issue I have, I hope they pass.

chriseppstein:
Hop into IRC today and we'll sort that out.

silvenon:
Btw, trig tests have already been fixed in [this](https://github.com/chriseppstein/compass/pull/734) pull request.

PR #761, opened 2012-03-08
Fix for Compass.shared_extension_paths when HOME is a relative path.

### LINKED ISSUE #364: Compass 0.11.1 no longer works for homeless root users
Deployed the latest Compass gem and got the following error upon starting my Resque worker:

```
rake aborted!
non-absolute home
/data/emaily/shared/bundle/ruby/1.9.1/gems/compass-0.11.1/lib/compass.rb:18:in `expand_path'
/data/emaily/shared/bundle/ruby/1.9.1/gems/compass-0.11.1/lib/compass.rb:18:in `shared_extension_paths'
/data/emaily/shared/bundle/ruby/1.9.1/gems/compass-0.11.1/lib/compass/configuration/helpers.rb:105:in `discover_extensions!'
/data/emaily/shared/bundle/ruby/1.9.1/gems/compass-0.11.1/lib/compass/app_integration/rails.rb:59:in `initialize!'
/data/emaily/shared/bundle/ruby/1.9.1/gems/compass-0.11.1/lib/compass/app_integration/rails/actionpack3/railtie.rb:43:in `block in <class:Railtie>'
```

The relevant code:

``` ruby
def shared_extension_paths
  [File.expand_path("~/.compass/extensions")]
end
```

The code is being run as root on a typically-configured Linux deploy target.

pivotal-casebook:
This should fully resolve https://github.com/chriseppstein/compass/issues/364 for when HOME is not just nil but also "." (or any other relative path). Added a complete set of unit tests around the method as well.

File.expand_path is unsafe to call if HOME is a relative path.

### DIFF
--- a/lib/compass.rb
+++ b/lib/compass.rb
@@ -16,7 +16,7 @@ module Compass
   end
   def shared_extension_paths
     @shared_extension_paths ||= begin
-      if ENV["HOME"] && File.directory?(ENV["HOME"])
+      if ENV["HOME"] && File.directory?(ENV["HOME"]) && ENV["HOME"].first == "/"
         [File.expand_path("~/.compass/extensions")]
       else
         []
--- a/test/units/compass_module_test.rb
+++ b/test/units/compass_module_test.rb
@@ -1,0 +1,36 @@
+require File.join(File.dirname(__FILE__), "..", "test_helper")
+
+class CompassModuleTest < Test::Unit::TestCase
+
+  def setup
+    Compass.reset_configuration!
+    Compass.instance_variable_set("@shared_extension_paths", nil)
+    @original_home = ENV["HOME"]
+  end
+
+  def teardown
+    ENV["HOME"] = @original_home
+    Compass.reset_configuration!
+  end
+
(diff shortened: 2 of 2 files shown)


lib/compass.rb
```diff
@@ -16,7 +16,7 @@ def lib_directory
   end
   def shared_extension_paths
     @shared_extension_paths ||= begin
-      if ENV["HOME"] && File.directory?(ENV["HOME"])
+      if ENV["HOME"] && File.directory?(ENV["HOME"]) && ENV["HOME"].first == "/"
```
chriseppstein, line 19: Pretty sure this breaks windows users because HOME there starts with `C:\` or something like that. I think a better approach is to rescue the `ArgumentError` that gets raised from `expand_path`.

pivotal-casebook, line 19: Yeah I also noticed that .first works when its loaded in Rails but .first is not defined when running compass. Need to use [0] in this case.

I'll rewrite with that.

pivotal-casebook:
Closing in favor of https://github.com/chriseppstein/compass/pull/763

PR #764, opened 2012-03-09
Add Thin and Foreman for doc-src

silvenon:
I (think I) made the Serve method faster with Thin. The Serve method is now primary, because the Nanoc method is really slow. I grouped those two Serve processes into a Procfile using Foreman, because the command `./bin/foreman start` is easier to remember and because I find it difficult to stop background processes, in this case `./bin/serve 300 output &`. (The Foreman version doesn't have the ampersand.)

### DIFF
--- a/doc-src/Gemfile
+++ b/doc-src/Gemfile
@@ -1,5 +1,6 @@
 source :gemcutter
 
+gem 'foreman'
 gem 'nanoc3'
 gem 'thin'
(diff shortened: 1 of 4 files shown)


silvenon:
Actually, this is a stupid pull request because the watch component doesn't work properly anyways. I'll be back if I manage to fix the environment for generating and previewing.

PR #768, opened 2012-03-11
Optimized property-heavy mixins by having them extend a class

anthonyshort:
This is a method that I'm currently using in Stitch to reduce the total weight of the stylesheets. Basically, creating classes for properties that are using in frequently used mixins, like clearfix can reduce the bloat of a stylesheet a great deal. 

You get the same interface of including a mixin so it doesn't require the user to create the class themselves. However, the mixin itself could be extended to allow the user to choose the class that is extended. (If it's even possible in Sass).

An added benefit to this is debugging. When inspecting the element the properties used for clearfixing are all grouped together. This means you won't see an overflow hidden in your property list for that selected element and not know what it's there for.

I'd like to see what other people think of this approach. As I said, I've been using this with Stitch for a while now with a number of mixins with no real drawbacks. So it could easily be extended to the other work-horse mixins of the framework. 

Although one drawback maybe that as soon as you import compass into the stylesheet you'll be adding these classes to the users stylesheet. Not a big deal really (for me), but it might have people complaining about compass bloating their stylesheets (which is ironic as it will actually reduce the total weight of the file in the end).

### DIFF
--- a/frameworks/compass/stylesheets/compass/utilities/general/_clearfix.scss
+++ b/frameworks/compass/stylesheets/compass/utilities/general/_clearfix.scss
@@ -10,6 +10,9 @@
 // Recommendations include using this in conjunction with a width.
 // Credit: [quirksmode.org](http://www.quirksmode.org/blog/archives/2005/03/clearing_floats.html)
 @mixin clearfix {
+  @extend .__compass-clearfix;
+}
+.__compass-clearfix {
   overflow: hidden;
   @include has-layout;
 }
@@ -19,6 +22,9 @@
 // has the advantage of allowing positioned elements to hang
 // outside the bounds of the container at the expense of more tricky CSS.
 @mixin legacy-pie-clearfix {
+  @extend .__compass-legacy-pie-clearfix;
+}
+.__compass-legacy-pie-clearfix {
   &:after {
     content    : "\0020";
     display    : block;
@@ -35,6 +41,9 @@
 //
 // Adapted from: [A new micro clearfix hack](http://nicolasgallagher.com/micro-clearfix-hack/)
 @mixin pie-clearfix {
+  @extend .__compass-pie-clearfix;
+}
+.__compass-pie-clearfix {
   &:after {
     content: "";
     display: table;


barraponto:
@anthonyshort I like your __class-name, I bet it avoids clashing with actual classes from the markup. I use -class-name in my stylesheets, but placeholder selectors are already coming our way in SASS 3.2. All you'll have to do is declare the placeholder class as %class-name and it won't print the placeholder selector at all.

I believe extending or including is up to the front end developer, though.

chriseppstein:
I _really_ want to do this in 0.13 on top of sass 3.2 which has the placeholder selector feature that @barraponto mentioned. I've been holding off on this feature because there's a stigma around css frameworks that they "force" you to use their selectors or give you "a bunch of stuff you don't need or use and make your stylesheets bloated". Placeholder selectors solve this problem perfectly.

Also, to do this right, I think we need to do a lot more than what is here.
1. Add documentation support for placeholder selectors so that each module provides a clear API of the base classes that it provides.
2. Keep the mixins as mixins and use them to define the placeholder selectors -- this will avoid document order/specifity changes.
3. Get people to start using @extend in their own projects by teaching them to use it in compass.

Other things to consider:
- In Sass 3.2 will can inspect the current selector context. We can use this to create a smart mixin/include system for mixins with small definitions that would do an include in the case where the selector is actually longer than the contents.

barraponto:
I guess it's also interesting to see that css3/pie already implements this behavior although in a less intrusive way: http://compass-style.org/reference/compass/css3/pie/#mixin-pie

anthonyshort:
I'll scrap this for now and take another look at it with placeholder selectors and other 3.2 goodies so we can do it right.

MoOx:
I want to share what I've done using placeholder selector feature

https://github.com/MoOx/compass-recipes/blob/master/stylesheets/recipes/shared/_clearfix.scss

I don't like the simple clearfix using overflow hidden (because for example, box-shadow is not working !), & with this nice approch with a combo of "`@include` using `@extend %`" it's transparent for the user.

Edit: I just did a refactor of my code

PR #769, opened 2012-03-11
Replaced all references to "icons" in the tutorial with "my-icons" to ma...

pietschy:
...ke it clearer that the @include all-icon-sprites is dependent on the directory name.

### DIFF
--- a/doc-src/content/help/tutorials/spriting.markdown
+++ b/doc-src/content/help/tutorials/spriting.markdown
@@ -15,33 +15,33 @@ of several convenient ways.
 
 For this tutorial, let's imagine that in your project's image folder there are four icons:
 
-* `images/icon/new.png`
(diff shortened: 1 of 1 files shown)


chriseppstein:
I had some merge conflicts and then took some of your changes a bit further. It's on stable now: 70423373935db54deee49f8c38caf8baf950c57d

pietschy:
Excellent, thanks.

I did think about adding a note or section about nested directories but was in a bit of a hurry.  I still think it might be worth having a bit of info on this.  I'm happy to have a go, do you have any conventions/styles for callouts/notes etc?

chriseppstein:
We did a lot of updates to the spriting docs for the 0.12 release. you should re-read them and make sure that's not covered. re: notes, not really, just whatever seems to flow with the content.

PR #776, opened 2012-03-13
Fix code style in vertical_rhythm's apply-side-rhythm-border().

JohnAlbin:
This is just a minor code style fix. There should be a newline before the closing } bracket.


scottdavis:
+1

PR #778, opened 2012-03-13
Rename $leader variable in vertical_rhythm's leader() and trailer() mixins

JohnAlbin:
`trailer()` appears to be a near copy-and-paste of `leader()`. Here it is:

```
@mixin leader($lines: 1, $font-size: $base-font-size, $property: margin) {
  $leader: rhythm($lines, $font-size);
  @if unit($leader) == px {
    $leader: floor($leader)
  }
  #{$property}-top: $leader;
}
```

But the internal variable `$leader` makes no sense inside `trailer()`. “WTF? Why is there a `$leader` variable inside the `trailer()` mixin?”

If we rename this variable to `$whitespace`, it makes sense in both mixins.

Alternatives:
1. we could just rename the `$leader` variable inside the `trailer()` to be `$trailer`.
2. we could write a `whitespace()` helper mixin that has a `$direction` parameter. And use that internally.

I actually prefer alternative 2 over the commit in this pull request, but it's a bit more radical of a change.

### DIFF
--- a/frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
+++ b/frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
@@ -113,6 +113,10 @@ $base-half-leader: $base-leader / 2;
     @warn "$relative-font-sizing is false but a relative font size was passed to the rhythm function";
   }
   $rhythm: $font-unit * $lines * $base-line-height / $font-size;
+  // Round the pixels down to nearest integer.
+  @if unit($rhythm) == px {
+    $rhythm: floor($rhythm);
+  }
   @return $rhythm;
 }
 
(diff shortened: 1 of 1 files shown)


JohnAlbin:
I uploaded the commit for alternative 2 so you could more easily see it as an option.

mirisuzanne:
I don't see the need for the new mixin. Is it really doing something that can't simply be done by the already-existing `rhythm()` function? Why not move the `floor()` trick over into `rhythm()` and just use that? All we're really doing is a bit of math, so we should return it as a value, not a property-value pair. Leave the properties in their specific mixins.

scottdavis:
Im for making whitespace a function to dry this up.

JohnAlbin:
A function? Hmm… let me give that a spin.

JohnAlbin:
I gave @ericam's suggestion a spin in the commit above.

mirisuzanne:
Looks good to me. Thanks!

PR #779, opened 2012-03-13
Add "css" keywords on the Compass Sprite helpers doc.

MoOx:
Following this (old) tweet https://twitter.com/#!/chriseppstein/status/141632327565578240
I do some (stupid) adds. We can try these and see what happen ?


MoOx:
Maybe we can add a paragraphe using words like "generator" "easy" "fast".
I do that because [this](http://wearekiss.com/spritepad) kind of "tools" piss me of.

scottdavis:
this wil go up the next time the docs are pushed

PR #782, opened 2012-03-13
Add ability to append stuffs after font urls

MoOx:
I wasn't able to use a [tricky font face syntax](http://www.fontspring.com/blog/the-new-bulletproof-font-face-syntax)
So I change how things works + add appropriate tests.
- Update outdated doc about font face.
  I hope I do it the right way.

### DIFF
--- a/test/units/sass_extensions_test.rb
+++ b/test/units/sass_extensions_test.rb
@@ -134,6 +134,10 @@ class SassExtensionsTest < Test::Unit::TestCase
       evaluate("font-files('/font/name.eot?#iefix')")
     end
 
+    assert_nothing_raised Sass::SyntaxError do
+      evaluate("font-files('/font/name.svg?mightbedynamic=something%20+escaped#fontId')")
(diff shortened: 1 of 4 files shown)


test/units/sass_extensions_test.rb
```diff
@@ -126,6 +126,14 @@ def test_font_files
...
     end
 
+    assert_nothing_raised Sass::SyntaxError do
+      evaluate("font-files('/font/name.svg#fontId')")
+    end
+
+    assert_nothing_raised Sass::SyntaxError do
+      evaluate("font-files('/font/name.eot?#iefix')")
+    end
+
```
chriseppstein, line 136: Can you add one more test and make sure it passes:

```
assert_nothing_raised Sass::SyntaxError do
  evaluate("font-files('/font/name.svg?mightbedynamic=something%20+escaped#fontId')")
end
```

MoOx:
Done :)

PR #783, opened 2012-03-14
Support css3 input placeholder

### LINKED ISSUE #418: Mixin for input placeholder
At least 4 browser now partially support  input placeholder style (firefox 4, safari, chrome, opera 11). It is reasonable to include experimental mixin for this.

http://stackoverflow.com/questions/2610497/change-an-inputs-html5-placeholder-color-with-css

ry5n:
Add +input-placeholder to style the html5 input pseudo-element in Mozilla, Webkit and IE10. Uses the @content feature of Sass.

Add mixin +input-placeholder-color($color) as shortcut to change the colour of the input placeholder.

Proposal to close #418.

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_input-placeholder.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_input-placeholder.scss
@@ -5,9 +5,9 @@
 // 1. Webkit and Gecko (Mozilla) already support placeholders (2012/Mar/13)
 // 2. Future-proofing for IE10: http://msdn.microsoft.com/library/ie/hh772745.aspx
 @mixin input-placeholder {
-	&::-webkit-input-placeholder { @content; }	// [1]
-	&:-moz-placeholder { @content; }						// [1]
-	&:-ms-input-placeholder { @content; }				// [2]
+	&::-webkit-input-placeholder { @content; }  // [1]
+	&:-moz-placeholder { @content; }            // [1]
+	&:-ms-input-placeholder { @content; }       // [2]
 }
 
 // Shortcut to set the color of html5 input placeholder text.
--- a/frameworks/compass/stylesheets/compass/css3/_input-placeholder.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_input-placeholder.scss
@@ -5,9 +5,15 @@
 // 1. Webkit and Gecko (Mozilla) already support placeholders (2012/Mar/13)
 // 2. Future-proofing for IE10: http://msdn.microsoft.com/library/ie/hh772745.aspx
 @mixin input-placeholder {
-	&::-webkit-input-placeholder { @content; }  // [1]
-	&:-moz-placeholder { @content; }            // [1]
-	&:-ms-input-placeholder { @content; }       // [2]
+	@if $experimental-support-for-webkit {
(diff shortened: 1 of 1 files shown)


mirisuzanne:
Interesting. Is there no official placeholder syntax? It would be nice to future-proof in that direction as well, if we could.

chriseppstein:
@ericam it is not part of any draft specification yet. But there are three conforming implementations so it will become a [de facto standard](http://en.wikipedia.org/wiki/De_facto_standard)

mirisuzanne:
@chriseppstein I'm just curious what name it will land with, between 'placeholder' and 'input-placeholder'. do we want to make a guess on that, attempting to future-proof against an official format?

ry5n:
@ericam I'm not aware of an official syntax yet, which makes me think I'm probably wrong… a quick web search implies that no, there's still no official syntax.

@chriseppstein You're right, those might be common style choices, although the size and family would be inherited and I imagine most authors wouldn't want to change them. Font-style might be more useful and more common. So, I'm thinking I'll change +input-paceholder-color to +input-placeholder-text([$color], [$style], [$family], [$size]) with args ordered most-common to least.

Of course I'm open to other suggestions.

ry5n:
Or maybe bypass the complexity and simply not include the shortcut mixin. This is an experimental feature after all, and @ content isn't that hard to use.

chriseppstein:
@ry5n I agree. Let's bypass the shortcut mixin for now. I like the transparency and simplicity of mixin content here.

@ericam Let's wait to see how the spec shapes up. Once there's a draft proposal we can add the official syntax.

chriseppstein:
Can you move this to the user_interface module? Also can you add tests and a documentation example?

For example:
- Opacity example [page](https://github.com/chriseppstein/compass/blob/stable/doc-src/content/examples/compass/css3/opacity.haml) and [content](https://github.com/chriseppstein/compass/tree/stable/doc-src/content/examples/compass/css3/opacity) becomes [this in the documentation](http://compass-style.org/examples/compass/css3/opacity/)
- and opacity is tested like this: [input file](https://github.com/chriseppstein/compass/blob/stable/test/fixtures/stylesheets/compass/sass/opacity.scss) and [output file](https://github.com/chriseppstein/compass/blob/stable/test/fixtures/stylesheets/compass/css/opacity.css)

ry5n:
Sure thing; will be a few days before I can get to it but should have it done over the weekend.

mirisuzanne:
That all sounds good to me.

ry5n:
@chriseppstein One thing though – are there any other mixins that use @content that I can look at to see how they're documented?

chriseppstein:
@ry5n No, You're a trailblazer so far :)

ry5n:
I've moved the +input-placeholder mixin to the _user-interface partial and written reference, example docs and a test. Unfortunately due to a Git mishap the changes are on [a new branch](https://github.com/ry5n/compass/commit/6a4dbc60433a05f426268885f93f478d9b595c0f). Can/should I close this pull request and open a new one from the new branch?

chriseppstein:
Yes that's fine.

ry5n:
OK, done. Closing in favour of https://github.com/chriseppstein/compass/pull/799

PR #786, opened 2012-03-14
Replace confusing $ie-font-ratio variable with $browser-default-font-size in vertical_rhythm partial

JohnAlbin:
This commit is for the discussion in #766. Comments should go in that discussion thread, not this one.

### DIFF
--- a/frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
+++ b/frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
@@ -9,8 +9,8 @@ $base-line-height: 24px !default;
 // Set the default border style for rhythm borders.
 $default-rhythm-border-style: solid !default;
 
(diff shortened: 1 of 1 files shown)


mirisuzanne:
+1

chriseppstein:
Regarding the variable. I doubt anyone is using it so it's fine if it goes w/o a deprecation (and it's easy for them to fix it if it is in use). However, I'd like to understand what are the ramifications of moving the styles to the `html` element to existing users. This aspect seems like it deserves a changelog entry at least explaining that these styles moved and why. Thoughts?

scottdavis:
i agree and  i just noticed that when i viewed the diff this needs some kind if description

chriseppstein:
I decided to land this on stable. Seems harmless enough after thinking about it more carefully. Merged as: 48bfe186afb871139577303626e6decd8726c9b6

PR #790, opened 2012-03-14
Deprecate reset-baseline mixin from vertical-rhythm partial

JohnAlbin:
Follow-up to #780 to deprecate reset-baseline() mixin.

### DIFF
--- a/frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
+++ b/frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
@@ -67,8 +67,9 @@ $base-half-leader: $base-leader / 2;
   }
 }
 
-// resets the baseline to 1 leading unit
+// @deprecated Resets the baseline to 1 leading unit.
 @mixin reset-baseline {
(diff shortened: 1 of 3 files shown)


chriseppstein:
@JohnAlbin Thanks for this patch. We need to add an `@warn` to the mixin as well so that users who are using it are made aware. Here's an example of this pattern in action: https://github.com/chriseppstein/compass/blob/stable/frameworks/compass/stylesheets/compass/css3/_box-shadow.scss#L63

scottdavis:
whats going on with this? :bump:

JohnAlbin:
Sorry. I was at a conference for 10 days and on vacation for a week.

I just pushed a new commit. Does that `@warn` text in that commit look okay?

JohnAlbin:
Sorry, pushed wrong commit before. I've updated the commit properly and then rebased it against master.

Warning says:

> This mixin has been deprecated and will be removed in the next version. Use adjust-leading-to(1) instead.

scottdavis:
@ericam can you verify that this went in with the big VR patch?

scottdavis:
Its depricated

PR #791, opened 2012-03-14
Update docs for reset-baseline() in vertical_rhythm partial

JohnAlbin:
Over in #780 the decision was made to deprecate the reset-baseline() partial in the master branch that will be Compass 0.13 (which I did in #790).

But we still need to better document that mixin in the stable 0.12 branch.

Here's a better description of the mixin with a mild warning about it not working in all circumstances.

### DIFF
--- a/frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
+++ b/frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
@@ -64,7 +64,8 @@ $base-half-leader: $base-leader / 2;
   }
 }
 
-// resets the baseline to 1 leading unit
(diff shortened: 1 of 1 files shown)


ry5n:
@JohnAlbin +1, although I'd make it clear why it won't work and when: “does not work on elements whose font-size is different from $base-font-size”

scottdavis:
Go ahead and fix the typo and just make it `@include adjust-leading-to(1, $base-font-size);` and put a `@warn` in there and ill merge this in.

chriseppstein:
I took @ry5n's feedback into account -- this is on stable now as 3f96108688c0456eceea02ba724caf9aff740c0a and 1aca9604015ea4d04ed52dacd5580e1c9c17a072

PR #792, opened 2012-03-15
Updated legacy staticmatic link

gilbert:
[Staticmatic](https://github.com/mindeavor/staticmatic2) supports compass 0.12.x by default, so no extra configuration is needed.


scottdavis:
will go live with next doc push

PR #801, opened 2012-03-19
Include Mozilla specific rules only if $experimental-support-for-mozilla set to true. Fixes #794

### LINKED ISSUE #794: inline-block mixin (and others) should also respect "$experimental-support-for-..." settings
Right now even if 

$experimental-support-for-mozilla: false

the mixin produces -moz specific rules such as:

display: -moz-inline-box;
-moz-box-orient: vertical;

okonet:
Please include my patch for the problem described in #794

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_inline-block.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_inline-block.scss
@@ -6,8 +6,10 @@
   @if $legacy-support-for-ie {
     & { *display: inline; }
   }
-  display: -moz-inline-box;
-  -moz-box-orient: vertical;
+  @if $experimental-support-for-mozilla {
+    display: -moz-inline-box;
+    -moz-box-orient: vertical;
+  }
   display: inline-block;
   vertical-align: middle;
   @if $legacy-support-for-ie {


mirisuzanne:
All that junk was there for old (now quite old) versions of Mozilla. It's really an issue of legacy support, not experimental support. I've stopped using this mixin, because it now messes with more than it fixes. I'm in favor of deprecating it rather than changing it.

scottdavis:
so do we want to close this and deprecate the mixin or merge this then deprecate?

chriseppstein:
I agree this is legacy support and not experimental support and so it should be governed by a different configuration option. Note how I do this for IE: https://github.com/chriseppstein/compass/blob/stable/frameworks/compass/stylesheets/compass/_support.scss#L1-17

mirisuzanne:
Can we get the change made from experimental-support to legacy-support, along with a deprecation warning?

chriseppstein:
I added a `$legacy-support-for-mozilla` variable and I only emit these properties if it is true. See commits 50d3006d1d512d4b29b57bcf78228812ce5eea4b and 84e9a684b9697d728a37abb14cb0aae2c4d2a790.

I also provide a config option to change the value of vertical-align or to disable it completely.

Closing this PR.

PR #808, opened 2012-03-24
Remove unused -o-border-radius and -ms-border-radius

joschka:
Neither Opera nor Microsoft ever implemented vendor prefixes for the border-radius property. So they can just be removed.


scottdavis:
@miketaylr is this correct? i don't use opera as much as a should

joschka:
for Opera see: http://my.opera.com/community/forums/topic.dml?id=651612
for MS should be clear: IE8 has no support for border-radius at all and IE9 supports border-radius without prefix

chriseppstein:
Seems legit.

miketaylr:
@scottdavis yep, we never had a `-o-border-radius`, afaik.

PR #816, opened 2012-03-28
Adds experimental CSS regions support

nearestnabors:
As seen in http://css-tricks.com/content-folding/ Only good for use w/
IE and Webkit right now. Too early for official implementation even.

### DIFF
--- a/doc-src/content/reference/compass/css3/regions.haml
+++ b/doc-src/content/reference/compass/css3/regions.haml
@@ -1,0 +1,15 @@
+--- 
+title: Compass CSS Regions
+crumb: CSS Regions
+framework: compass
+stylesheet: compass/css3/_regions.scss
+meta_description: Specify CSS Regions for supported browsers.
+layout: core
+classnames:
+  - reference
+  - core
+  - css3
+---
+- render 'reference' do
+  %p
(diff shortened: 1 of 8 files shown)


chriseppstein:
Thank you for the patch. I would like to add this feature.

Can you please add docs and tests for it so that other users can know it exists and how to use it and we can make sure it doesn't break going forward?

Here's our [contribution guide](http://compass-style.org/help/tutorials/contributing/)

Thanks!

doc-src/content/reference/compass/css3/regions.haml
```diff
@@ -0,0 +1,15 @@
...
+meta_description: Specify CSS Regions for supported browsers.
+layout: core
+classnames:
+  - reference
+  - core
+  - css3
+---
+- render 'reference' do
+  %p
+    Provides a mixin for the CSS regions, properties which allow you to flow content into new containers. See <a href="http://dev.w3.org/csswg/css3-regions/">the spec draft</a> and <a href="http://labs.adobe.com/technologies/cssregions/">Adobe's page on the topic</a>.
```
chriseppstein: it actually provides several mixins :)

PR #819, opened 2012-03-29
Adds normalize.css framework.

ghost:
This is a plain copy of [normalize.css](https://github.com/necolas/normalize.css/). It would be great to have this as an alternative to a css reset.


scottdavis:
I would actually prefer something like this remain an extension for maintainablity sake

mirisuzanne:
+1 for a plugin

ghost:
Alright then, here we go: [https://github.com/jroettger/compass-normalize-plugin](compass-normalize-plugin). I am still working on the documentation, but it's working and everything.

scottdavis:
cool

PR #838, opened 2012-04-08
Border radius prefixes

ai:
Opera and IE support `border-radius` without vendor prefixes, so `-o-border-radius` and `-ms-border-radius` doesn’t exists. Proof link: http://blog.danielfriesen.name/2011/01/05/css3-border-radius-and-vendor-prefix-fud/.


ai:
Sorry, wrong branch to merge.

Issue #843, opened 2012-04-11
vertical-rhythm module should allow for true relative font sizing


mirisuzanne:
The current master branch version of `establish-baseline` works as follows:

``` scss
// input
@include establish-baseline(14px); 

// output
* html {
  font-size: 87.5%; }

html {
  font-size: 14px;
  line-height: 1.143em; }
```

The percentage-based output is hidden behind a hack, so it is only seen by IE6/7. All other browsers get a font-size set in px. While that seems to be accepted practice, it breaks the ability for browser users to set their own default font size. This becomes even more important with media-queries, where the only value of "em" that is recognized comes directly from that browser default setting. The only way to create a meaningful em-based media-query is to know the relative value of your own ems. The only way to know that is by setting the one in relation to the other.

PR #844, opened 2012-04-11
establish relative font size when $relative-font-size

mirisuzanne:
Fixes Issue #843

### DIFF
--- a/frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
+++ b/frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
@@ -55,14 +55,16 @@ $base-half-leader: $base-leader / 2;
 
 // Establishes a font baseline for the given font-size.
 @mixin establish-baseline($font-size: $base-font-size) {
-  // IE 6 refuses to resize fonts set in pixels and it weirdly resizes fonts
-  // whose root is set in ems. So we set the root font size in percentages of
-  // the default font size.
-  * html {
-    font-size: 100% * ($font-size / $browser-default-font-size);
+  $relative-size  : 100% * ($font-size / $browser-default-font-size);
+
+  @if $legacy-support-for-ie6 and (not $relative-font-sizing) {
+    // IE 6 refuses to resize fonts set in pixels and it weirdly resizes fonts
+    // whose root is set in ems. So we set the root font size in percentages of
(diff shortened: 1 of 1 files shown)


JohnAlbin:
I'm in favor of this change.

This _will_ change how the CSS gets rendered by default. Since the commit is re-using the `$relative-font-sizing` variable, that's inevitable. `$relative-font-sizing` will now actually respect the user's default font size. I think that is actually the intent of authors who _choose_ $relative-font-sizing to be true. Authors who don't notice the default and winded up with a font-size set in `px` in the html element and `em`s everywhere else… they basically got a _fixed_ font-size set with `em`s.

To make it clearer, here's the scenarios I see:
1. author wants relative fonts on the base element so the user's default font size is respected. This person is currently SOL because `establish-baseline()` currently sets a `16px` fixed font size.
2. author wants to override the user's choice with the designer's font size.
3. author just uses Compass mixins and defaults because Compass is awesome, but hasn't really thought about 1 or 2.

Currently #2 and #3 are satisfied by the default values. But they have very different desires. The people in #2 should be okay if they change `$relative-font-sizing` to false. Though that means they need to update their project's sass when the new version of Compass comes out.

The people in #3 might be surprised by this commit because they may stumble upon users who have set a different default font size in their browser.

I guess my only comment on this commit is that we _may_ want to set the default value of `$relative-font-sizing` to `false`. Though best practice really should be to set it to `true`.

JohnAlbin:
I'd like to point out, I consider this a bugfix. Though that won't make the people in scenario # 2 less upset that they have to update their Sass.

mirisuzanne:
Agreed. It changes the output for most users, so we have to acknowledge that. But I think it's the way things ought to work.

Is anyone opposed?

ry5n:
Not opposed. This is how it should work.

As for the default value, changing it may be /more/ disruptive. AFAICT if it were defaulted to false, people in #3 (certainly the majority of users) who update Compass will not only get pixel output instead of ems, they will get compile warnings where they had supplied a context font-size (mixins will @ warn `@ if not $relative-font-sizing and $from-size != $base-font-size`).

mirisuzanne:
Yeah, I'd vote for leaving the default. This seems the most sensible way to understand what the variables actually mean, and it's a change that wont affect most people in any noticeable way. It should simply get documented in the changelog. I'll merge this change, and then make that change as well.

PR #847, opened 2012-04-13
Turn off $legacy-support-for-ie* for @content within the @keyframes mixin.

StanAngeloff:
If a project has legacy IE support turned on, certain mixins add property
definitions which are not affected by $experimental-support-for-*. One
such mixin is @opacity with will add a `filter:` property.

Because IE 6, 7 and 8 do not support CSS animations and keyframes, we can
safely turn off the support and restore to its previous values once we are
done rendering the @keyframes @content.


mirisuzanne:
Good catch. Thanks.

PR #848, opened 2012-04-14
Add $offset parameter to rhythm() function.

JohnAlbin:
I was writing a mixin that needed to subtract a couple pixels from the vertical rhythm margin, just like the rhythm borders mixins do.

And then I noticed the math the rhythm border mixins are using is _almost_ identical to the math in the `rhythm()` function.

Keeping with DRY, I'd like to propose we add an $offset parameter to the `rhythm()` function.

Then I can use `rhythm()` in my mixin instead of copy/pasting/rewriting lines of code from the rhythm borders mixins.

Note that the $offset parameter is totally backwards compatible with existing usage of the `rhythm()` function. No one's existing Sass should be affected.

### DIFF
--- a/frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
+++ b/frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
@@ -111,12 +111,13 @@ $base-half-leader: $base-leader / 2;
 // Calculate rhythm units.
 @function rhythm(
   $lines: 1,
-  $font-size: $base-font-size
+  $font-size: $base-font-size,
+  $offset: 0
 ) {
   @if not $relative-font-sizing and $font-size != $base-font-size {
     @warn "$relative-font-sizing is false but a relative font size was passed to the rhythm function";
   }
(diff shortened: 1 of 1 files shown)


mirisuzanne:
Love it. +1

ry5n:
Awesome! Would be handy in other cases, eg. creating a mixin that shifts an element 'out of phase' (like a heading with more space above and correspondingly less space below) while maintaining the rhythm.

frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
```diff
@@ -112,11 +112,12 @@ $base-half-leader: $base-leader / 2;
 @function rhythm(
   $lines: 1,
   $font-size: $base-font-size
```
mirisuzanne: missing a comma at the end of this line.

JohnAlbin:
And I added a semi-colon after `$offset: 0`! Not sure how I managed both of those errors. :-p

Fixed them up and pushed a new commit.

PR #855, opened 2012-04-20
Multiple image dirs

buhrmi:
Support for multiple image dirs.

When working with Rails > 3.1, images are located at different locations. (app/assets, vendor/assets, lib/assets, you name it). So it would be nice if compass supports multiple image dirs when using the image dimension helpers.

This branch goes hand-in-hand with the the multiple_image_dirs branch for compass-rails: https://github.com/buhrmi/compass-rails/commit/fee5b8ab525951438196edf516fc601fd0f6861f


buhrmi:
some bugs appeared that have not been covered by the tests. will do some more work on it.

PR #856, opened 2012-04-23
Multiple image dirs

buhrmi:
This branchs makes the images_dir config option take an array of directories in addition to a string. basically to make it easier to work together with asset pipelines that pull images from different locations upon deploy. Goes hand in hand with https://github.com/Compass/compass-rails/pull/38

### DIFF
--- a/doc-src/content/reference/compass/helpers/image-dimensions.haml
+++ b/doc-src/content/reference/compass/helpers/image-dimensions.haml
@@ -20,8 +20,8 @@ documented_functions:
       image-width(<span class="arg">$image</span>)
   .details
     %p
-      Returns the width of the image found at the path supplied by <code>$image</code>
-      relative to your project's images directory.
+      Returns the width of the first image found at the path supplied by <code>$image</code>
+      relative to your project's image directories.
 
 #image-height.helper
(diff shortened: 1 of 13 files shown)


chriseppstein:
I think if we're going to add multiple directory support it needs to be:
1. for all the asset properties.
2. against master -- this is a big change. I want to think very carefully how this impacts our config, plugin api, etc.

geekdave:
Just wondering if any progress/decisions had been made on this issue.  I'm working on a large multi-module enterprise application built on Foundation/Compass.  We have separate directories for our production code, and our 3rd party libraries.  We'd also like to have a separation of our 3rd party assets (like images) from those designed in-house.  So having support for multiple images directories would be great.

scottdavis:
It's not something I'm actively focused on but I will consider a patch

stephenway:
@geekdave I have a similar setup at my work and the way I overcame this issue was by creating a private Compass extension specifically for themes and assets. This might not be the best solution for you, but I can show you what I've used to achieve this if it would help.

geekdave:
@stephenway : I'd very much appreciate that.  Thank you!

PR #863, opened 2012-04-26
Responsive Web Design helper mixins

Snugug:
I thought my [variable driven respond-to and media-query mixins](https://gist.github.com/2493551) felt like things that should belong in Compass for everyone to use, so here's a patch to add them to the Compass Layout partial plus a _rwd.scss partial to house the actual mixins.  Documented what they do/how they work in line, happy to document them in .md format. Obviously not useful until 3.2 drops, but I'm going to put this here so when it does we can have these in Compass.

### DIFF
--- a/frameworks/compass/stylesheets/compass/_layout.scss
+++ b/frameworks/compass/stylesheets/compass/_layout.scss
@@ -1,3 +1,4 @@
 @import "layout/grid-background";
 @import "layout/sticky-footer";
 @import "layout/stretching";
+@import "layout/rwd";
--- a/frameworks/compass/stylesheets/compass/layout/_rwd.scss
+++ b/frameworks/compass/stylesheets/compass/layout/_rwd.scss
@@ -1,0 +1,111 @@
+// Based on the [respond-to mixin in Sam Richard's Aura framework](http://github.com/Snugug/Aura)
+//
+// Mix into top of stylesheet
+//
+// Respond-to Usage:
+//  - Set named breakpoints in the $breakpoints variable
+//    - Assumes breakpoints are min-width and screen media queries
+//    - Minimum needed is a name and a value
+//    - Optional operator and query inputs for each breakpoint
+//    - If you want to change query, you need to include operator.
+//    - Format: 'name' value 'operator' 'query'
+//      - $breakpoints: 'foo' 500px, 'bar' 700px 'max-width', 'baz' 700px 'min-width' tv, 'qux' 900px 'max-width' tv;
+//        - Foo assumes the defaults of min-width and screen. Foo will generate @media screen and (min-width: 500px)
+//        - Bar changes min-width to max, but assumes screen as well. Bar will generate @media screen and (max-width: 700px)
+//        - Baz wants to use min-width, but change media type, so min-width needs to be re-declare. Baz will generate @media tv and (min-width: 700px)
+//        - Qux goes all out and has a full unique media query. Qux will generate @media tv and (max-width: 900px)
+//  - Input is name of named breakpoint.
+//  - Call the respond-to mixin with the named breakpoint you'd like
+//    - @include respond-to('foo');
+//
+// Media-query Usage:
+//  - Writes a media query based on input variables
+//  - Required input is $value
+//    - $value can either be a single value or a list of value, operator, and query
+//  - Optional inputs are $operator and $query
+//    - $operator defaults to 'min-width'
+//    - $query defaults to 'screen'
+//  - Call the media-query mixin with what you're querying/
+//    - @include media-query(300px);
+//    - @include media-query(300px, 'max-width');
+//    - @include media-query(300px, 'max-width', 'tv');
+
+
+// Defaulted variable for use with respond-to mixin
+$breakpoints: false !default;
+
+// Mixin to dynamically generate named breakpoints based on the $breakpoints variable
+@mixin respond-to($context) {
+  @if $breakpoints != false {
+    // Check to see if the 2nd item is a number. If it is, we've got a single query
+    @if type-of(nth($breakpoints, 2)) == 'number' {
+      // Check to see if the context matches the breakpoint namespace
+      @if $context == nth($breakpoints, 1) {
+        // Call Media Query Generator
+        @include mqg($breakpoints) {
+          @content;
+        }
+      }
+    }
+    // Else, loop over all of them
+    @else {
+      // Loop over each breakpoint and check context
+      @each $bkpt in $breakpoints {
+        // If context is correct…
+        @if $context == nth($bkpt, 1) {
+          // Call the generator!
+          @include mqg($bkpt) {
+            @content;
+          }
+        }
+      }
+    }
+  }
+}
+
+// This functionality gets used twice so I turned it into its own mixin.
+// Not really to be called on its own, used to slice out breakpoint name from input.
+@mixin mqg($bkpt) {
+  // Get length of breakpoint variable, minus the namespace
+  $length: length($bkpt);
+  // Go through all of the breakpoint items, starting at the second, and add them to a variable to be passed into the media query mixin
+  $mq: false !default;
+  @for $i from 2 through $length {
+    // If it's the first item, override $mq
+    @if $i == 2 {
+      $mq: nth($bkpt, $i);
+    }
+    // Else, join $mq
+    @else {
+      $mq: join($mq, nth($bkpt, $i));
+    }
+  }
+  // Call Media Query mixin
+  @include media-query($mq) {
+    @content;
+  }
+}
+
+// Slightly modified version of [Sam Richard's Media Query Mixin](https://gist.github.com/2490750), modified to accommodate list input
+@mixin media-query($value, $operator: 'min-width', $query: 'screen') {
+  // If a list is passed in for value, break it into value, operator, and query
+  @if type-of($value) == 'list' {
+    $mq: $value;
+    
+    @for $i from 1 through length($mq) {
+      @if $i == 1 {
+        $value: nth($mq, 1);
+      }
+      @else if $i == 2 {
+        $operator: nth($mq, 2);
+      }
+      @else if $i == 3 {
+        $query: nth($mq, 3);
+      }
+    }
+  }
+  
+  @media #{$query} and (#{$operator}: #{$value}) {
+    @content;
+  }
+}


chriseppstein:
I have concerns about creating a new syntax effectively representing media queries. It really feels like we just need a way to map names to canned queries.

I think @anthonyshort was working on something like this, I'd like to have him drive this feature.

anthonyshort:
I agree with Chris. A new pseudo-syntax probably isn't the way to go. 

We can leave it super flexible using a respond-to mixin with custom variables or create a bunch of mixins for canned queries that can be modified by variables.

This is the option I have been leaning towards as it's easiest for people to use and the semantics added to the stylesheet are great. Here's an example - https://gist.github.com/2028061

However, this isn't exactly future-proof. The most common tablet or phone size could change. We shouldn't be targeting devices when using these queries. I think we need something in between.

Snugug:
@canarymason has something built that I think is really good, gem install breakpoint I believe, that I think would be good for Compass.

I personally don't feel like canned media queries are a good idea unless the user is defining them as good responsive design is based on a site's content not on known device sizes. I can understand not wanting to redefine media queries through my full @respond-to mixin, but having an abstracted mixin like the media query mixin I wrote, or really what I consider the successor which is Mason's, is the better way to go and something that we should have in Conpass

Snugug:
To me, the solution linked to is something that isn't flexible enough and isn't future proof enough to be considered. Hell, it isn't even something that follows RWD best practices. Here's Mason's breakpoint mixin:

https://github.com/canarymason/breakpoint

I prefer connecting this to my respond-to solution, and then you can do whatever you want. 

If the options are between that linked solution and "redefining media queries" in such a way to be as flexible as possible while making them future proof and conform to best standards in RWD, I'll take the later.

anthonyshort:
Those are the same concerns I had and why we want to make sure we get it right.

Breakpoint is more along the lines of what I'd like to see included with a number of predefined variables included with compass rather than a pseudo-syntax. From there people can create their own abstracted mixins for specific devices if they want to have 'semantic' mixins. 

One thing I want to avoid is creating a solution that is extremely complex to the point where it would be easier to just write a couple of variables at the top of the stylesheets and use stand @media-query definitions using the variables, eg https://gist.github.com/2821753 - 

That is the problem I see with Breakpoint. It's easier to use standard media queries with variables, in my mind, as they're almost the same thing.

Another potential solution could be defining a list of points and having the rest taken care of by Compass.

```
$breakpoints: 200px 300px 500px 700px 900px;

@include breakpoint(1,3) { }
@include breakpoint(1) {}
@include breakpoint(1,5) {}
```

That is how designers would approach the situation. From there you can specific min and max points using the two parameters of the mixin.

Being future proof is great but I think a large part of the goal is to make everyday responsive patterns easier as writing them really isn't that hard.

chrisbuttery:
I like this approach of defining specific break points @anthonyshort, but if we're talking future proof, whats going to happen when a new breakpoint is added - say 255px. 
We'd then have to go through and update all previous includes again to reference its correct point.

Whats wrong with something like :
@include breakpoint(250px, 500px) { };

Its not dependent on any predefined variables.

Snugug:
I'm telling you that defining specific breakpoints is a really, really bad idea as it not only isn't future proof, but if you're doing RWD correctly, they also aren't going to mean anything as the breakpoints you're going to want to use are breakpoints based on your content and not on devices anyway.

Read http://snugug.com/musings/principles-responsive-web-design (yes it's my site, but it's the collective best practices for RWD, which is really what we're talking about). 

@anthonyshort, your concern with Breakpoint is unfounded as the most common breakpoint (size with min-width and screen) is as simple as $bps: size; @include breakpoint($bps) {}. It's where the more complicated breakpoints come in where you see more of it, but $bps2: 200px 300px 'height'; @include breakpoint($bps2){} is still ages and spaces better and easier than writing @media screen and (min-height: 200px) and (max-height: 300px). It's also a much better developer experience than @include breakpoint(1) or @include breakpoint(1,3) as every time you'd want to use a media query you'd need to reference that variable, and it's something that is future proof. 

@chris, to answer your question, that's precisely what @breakpoint does, except with a list as opposed to straight input. 

Anthony and other Chris, to sum up, a solution like Mason's @breakpoint is a better DX, more future proof, and more versatile than $breakpoints: a b c @include breakpoint(1, 3), and conforms to RWD best practices. Additionally, that solution, or similar solutions, are currently being used in production environments and are written, making integration easier than a new solution.

anthonyshort:
I'm well aware of responsive design best practice. I'm just trying to find a solution that is useful enough to be included. Can you give an example of how Breakpoint or your solution is easier or more beneficial than this - https://gist.github.com/2821753 ? 

I just don't see how `@include breakpoint($bps2)` or `$breakpoint: 200px 400px` is "ages and spaces" better than `@media $breakpoint-2 {}` or `$breakpoint: "screen and (min-width:200px) and (max-width:400px)";`

My main issue with breakpoint is that it's abstracting the syntax of media queries without much benefit other than slightly less characters. The media query syntax isn't hard or long so what's the point in abstracting it?

In practice many people just want to target the common devices. I'm not saying it's correct in terms of best practice. I'm just saying that in the real world most people just want an easier way to target phones and tablets without needing to remember the queries. 

If you're going to have breakpoints based on the content and forget about devices, great you're doing the right thing, but you will hardly need abstraction for that. Content-based breaks occur at different points for different elements on the page and in many cases they don't relate to each other. So the number of points could grow quite large and using separate variables might not even be useful.

If we want to conform to best practice we wouldn't need any abstraction at all. We have variables and we have nested media queries. What I think we need to achieve is a way of making common patterns easier for developers. 

What say you, @chriseppstein?.

Snugug:
$breakpoint: 300px is more beneficial than your gist as it doesn't require essentially the whole media query stack to be written and it abstracts out the most common usage and makes writing more complex ones easier, which is, after all, what a framework should do. It's also a much better DX than the proposed solution with +breakpoint(1, 3).

If your argument is that we should include device breakpoints in an easy to use manner but not a more robust system that, let's be honest, does the same thing but in a more future proof and overall useful way, I'd say let's not include a system at all, after all isn't +breakpoint(1, 3) just an easier way to write @media screen and (min-width: 320px) and (max-width: 480px), but with the disadvantage of it not future friendly and worthless for anyone doing RWD correctly? What makes +breakpoint(1, 3) more usable than +breakpoint($iphone_landscape)? Personally, the latter is much easier for me as a developer to use because I know what breakpoint I'm calling because of its semantics. 

As for, in practice people wanting to target devices, I can tell you that that's simply not true. In practice, what happens is 
you either have a good designer who tells the client it's not the way to do it, or they do and realize it doesn't work out the way they expect and start making content based changes, making what would be in Compass useless.  Additionally, do we, as a community, really want to be encouraging bad development practices by limiting our frameworks to design practices that are out of date, or do we want to include a system that is as future friendly as is at the heart of CSS Preprocessing itself?

anthonyshort:
You're taking my quick suggestion too seriously :) I thought of that while I was writing, it's in no way a well-thought-out solution. My point was that I think what abstraction we need, if any, is a quick way to define all the breakpoints in a stylesheet.

I know targeting devices doesn't work in a lot of cases. I'm just not convinced by the shorter syntax with either of the solutions you've proposed.

I do agree that including device mixins is probably not the best idea and we do need to promote best practices. People do target devices though, look at any of the vanilla CSS grid frameworks (http://twitter.github.com/bootstrap/scaffolding.html#responsive) they all do it. So yeah, it's true. People do it. 

I think our options are to either abstract the syntax out like you've suggest with Breakpoint or not doing it at all. Perhaps it all comes down to coding conventions rather than something that is able to be implemented by a framework. 

I'm not on any particular side, I see the benefits of every approach, I just want to find the right one for Compass. There is a point where head-in-the-clouds semantics falls apart but it's a slippery slope to go in the other direction.

codingdesigner:
The reason I wrote Breakpoint was that I wanted a very simple and easy to remember syntax for writing media queries. I often forget the standard syntax and need to look it up as I work. I had also written a few project specific "respond-to" mixins that worked by switching on a keyword variable. Those became difficult to manage and maintain as I added options. Each new option added a few more lines to read and manage. 

Since most of the queries I write (and most that I see in the wild) are for screen, and checking for a min-width I feel pretty comfortable setting those as defaults, which takes a lot of the thought out of writing the simple queries. It doesn't add new functionality above writing a straight query or query variable, but I like this and I think others will too. 

I like how this solution allows developers to assign a name to each breakpoint variable that maps to a semantic need or device. `$breakpoint-handheld-portrait: 320px 480px` is much easier to remember  than `$breakpoint-handheld-portrait: screen and (min-width:  320px) and (max-width: 480px);` It also allows you to manipulate it like standard Sass data, which you can't do with a string-based approach:

```
$breakpoint-handheld-portrait: max-width 320px;
$breakpoint-larger-than-handheld-portrait: nth($breakpoint-handheld-portrait, 2) + 1;
```

When used this way I think it could be the basis for a configurable rwd grid system that starts with common device breakpoints, but you'd override with your own values and variables. 

There's some work in breakpoint to address browser prefixing in device-pixel-ratio. Opera makes that pretty tricky right now, and it would be nice to address this in Compass the way so many other vendor prefixes are handled. (https://github.com/canarymason/breakpoint/blob/master/stylesheets/_breakpoint.scss#L57)

I also think it's good that you don't _need_ to set up variables for this to work. If someone prefers, they can simply write, `@include breakpoint(320px 480px) {}` and it works. They can refactor later. 

Of course tho, this is my baby and it's very new and cuddly to me. I see how cool it is and think it can be helpful to others. But it's not perfect (yet) and I understand your caution. No one wants feature bloat in compass. That's why we have extensions. If there's interest tho I'll go the next step and write a patch.

chriseppstein:
Ok, so I've read through this thread and all the links. that took quite a while :)

I don't have the experience that you two have with building responsive sites and so I can't approach this conversation from a best-practices perspective.

I would like to take a step back and ask what are the problems we are trying to solve by creating a RWD module for compass. 
- What are the pain points in day-to-day authoring?
- How can this module insulate against predictable change over time, project maturity, browsers & devices, etc.
- Are there some core concepts that compass can help users by forcing them to think them through or by abstracting them?
- How can compass make RWD accessible to users new to the practice?
- How can compass make RWD fast for experienced developers?

I agree that compass should not get into the business of providing predefined breakpoints or queries. Rather we should be in the business of helping people construct their own 

Additionally, the micro-syntaxes for media queries that I've seen so far do not really save much time nor do they improve the legibility over the existing MQ syntax. But they do make this module harder to learn and once learned, that knowledge is not portable. If we need to inspect the fine-structure of a media query, I would rather create an API that allows for the reading of values from a standard media query authored by the user than an API that produces media queries from a terse data structure.

So let's get the goals all out on the table so that proposed solutions can be evaluated against them.

MoOx:
Today for me the problem is we have to write a long line  for each mq. With Sass 3.2, the problem is almost solved since we can add variable. So If we add something to Compass for developers, it have to be clear & concise & semantics IMO.

I just throw some stuffs I'm working on https://gist.github.com/2833853

Simple usage must be something like

``` scss
.selector
{
    rule: stuff;

    @include media-mobile()
    {

    }

    @include media-tablet()
    {

    }

    @include media-desktop()
    {

    }
}
```

Maybe another way can be to use semantics using words small, medium, big or tall.

Does this really have to be futur proof since the web is moving so fast ?

Snugug:
Yes, it needs to be future proof as RWD exists in no small part to future proof our sites. 

The gist you linked to seems like a very complex way of essentially doing what Mason's Breakpoint does, with the disadvantage that it doesn't help write anything but those specific media queries and is much harder to follow. Also, the semantics of mobile, tablet, and desktop when dealing with media queries is not only a bad practice, both to get our users use to and as actual RWD best practices, but it's something that doesn't work in practice as a 1024x768 screen is just as easily a "tablet" as it is a "desktop", and 600x800 screen just as easily a "mobile" as it is a "tablet". Assuming context based on screen size is a classic mistake, and we shouldn't be delivering something to the community is built on something that is so easily identifiable as a bad practice.

As for the argument that Bootstrap or Foundation use device device breakpoints, I ask do we really want to be following what a group using Less is doing (I kid). Seriously though, those are frameworks designed with absolutely no knowledge of a user's content that needs to appeal to as wide an audience as possible, so their choice is to either use device breakpoints or to make up ones that they would then need to defend in another way. Just be because the big boys does it doesn't me it's what we should be doing; Twitter, Facebook, and Google UA sniff for mobile detection/redirection, and no one seriously doing RWD thinks that's a good idea.

In the end, I think we have a choice to make; either provide a new syntax that makes writing long media queries easier and makes writing short ones drop dead simple for any media query (like Breakpoint), or we don't include one at all and have solutions live in Extensions.

codingdesigner:
@chriseppstein I had some time on a long bus ride yesterday to consider your questions. Since I couldn't post from the bus, I also had time to format them. :)
- What are the pain points in day-to-day authoring?
  - query syntax
  - managing breakpoint values beyond a very limited set
  - seeing the relationship between the query and the reason for writing it
  - an api will ensure that 'and's and parentheses are always correct
- How can this module insulate against predictable change over time, project maturity, browsers & devices, etc.
  - vendor prefixing new features
  - managing queries as variables helps to manage a growing or aging project
- Are there some core concepts that compass can help users by forcing them to think them through or by abstracting them?
  - I'm not sure about those one. Your point about non-portable knowledge is well-taken. 
- How can compass make RWD accessible to users new to the practice?
  - reducing the syntax for common queries to a mixin with a single argument should help them get started
  - creating a set of breakpoint variables should help them keep track of why they need each query if they're named well
- How can compass make RWD fast for experienced developers?
  - semantically named variables will make development faster
  - being able to do math on the numbers in a query (`$small + 1`) will encourage systematic thinking

chriseppstein:
So if we want to encourage a mobile-first, content-first approach to RWD, we should make an API that forces the user to express their content's parameters instead of thinking about what breakpoints exist. Then breakpoints could be added and selected based on what media queries are defined. If more breakpoints are added, the media query selected might change.

Something like (I'm spit-balling here):

``` scss
$screen-breakpoints: "min-device-width: 320px", "min-device-width: 480px", "min-device-width: 960px", "min-device-width: 1280px";
#sidebar {
  @include when-wider-than(300px) { float:left; }
  @include when-wider-than(700px) { position: fixed; left: 50px; top: 100px; }
}
```

Could produce:

``` css
@media screen and (min-device-width: 320px) { #sidebar { float:left; } }
@media screen and (min-device-width: 960px) { #sidebar { position: fixed; left: 50px; top: 100px; } }
```

chriseppstein:
In terms of configuration. It could be more like:

``` scss
$rwd-screen-width: device; // defaults to normal widths instead of device widths?
$rwd-screen-breakpoints: 320px, 480px, 960px, 1280px;
```

Also, please keep in mind that Sass media queries, when nested, will self-assemble. So we could have the following mixins:
- `when-wider-than($width, $until: false)`
- `when-oriented($orientation)`
- `when-pixel-ratio($ratio)`
- `when-aspect-reaches($horizontal, $vertical)`

and by simply nesting these mixins, you'd get the AND and parens added by Sass itself.

Snugug:
I forgot about the self assembling, that will make the breakpoint mixin easier to maintain. 

That being said, I don't think that having mixins for each type of media query is very sustainable. There are, as of the Level 4 spec, 16 official media features that can be queried off of, with something like 5 or 6 (relevant) vendor specific ones used widely, with some of those being prefixed and implemented different (very specifically, device-pixel-ratio), with 9 official ones allowing for min and max options, with 10 different media types that can be accounted for. This, if my math is to be trusted, comes out to be about 400 different mixins we would need to write to accommodate every permutation of single-serving media queries that could be written. Even if we were to go with the most common ones, we would still need approximately between 81 and 108 (if we literally only include 'all' and 'screen' as options, 54) different mixins, and that's still not quite future proof.

With all those numbers crunched and with that being said, I firmly believe that if we are going to include MQ support in Compass, it should be a single mixin that makes smart choices about default input the way Breakpoint is handling it. True, the mixin name isn't as syntactically nice as 'when-wider'than', but RWD has gotten the idea of breakpoints and media queries so firmly in our heads that I think that's OK. Those two words mean something to developers, whereas 'when-wider-than' will get tiresome to write for experienced developers and moves new developers even further away from the base idea we are working from.

chriseppstein:
No, if we followed this approach, we'd need, at most, one mixin for each media feature, min, max, etc would be handled by arguments.

But I think we differ on something very substantial here. It seems that you feel we should either implement an API that fully expresses the media query syntax, or we should do nothing. And I just don't agree with this sentiment. The users can build their own mixins and use Sass 3.2's very robust media handling to achieve the full suite of media capabilities. 

I think compass can add value in a RWD module by encapsulating the best practices in a way that makes it trivial to follow them and by using good naming and a constrained approach set the mental model for the user. A good analog is Grids and Vertical rhythms. These enforce constraints that are derived from common usage patterns and make it clear what following the rules looks like and while breaking them is allowed, it also obvious to someone who is trained that the rules were broken (E.g. you never set width or line-height directly).

In terms of future proofing, if a new aspect is needed or if the API doesn't quite meet a user's need, well they just drop an @media query in anywhere and Sass will bubble it and combine it with our existing mixins.

But I really can't speak to what the common usage patterns for RWD are, I could certainly research it, but I think that all of you are smart and well travelled developers and I would like to lean on your insight. So for the time being, stop trying to make everything possible and start thinking about how you can make the 90% common case as optimal as possible while simultaneously discouraging the kind of device centric approach that we know will be brittle over time. Maybe my example isn't what it should be -- so tell me what it should be.

chriseppstein:
@canarymason Thank  you for breaking that down. It's helpful, but I leaves me with more questions than answers. Can you provide some code examples or further details that exemplify the common problems? For instance "query syntax" -- what about the query syntax is difficult? "managing breakpoint values beyond a very limited set" -- what is a real world example of a complex set of breakpoint values and what aspects of them became difficult to manage.

Snugug:
This comment is reposted below with formatting:

chriseppstein:
reposting your comment to see if it formats correctly:

For me, the most common media queries are as follows(in approx order; CSS4 queries have *; not yet implemented, projected usage) 

```
screen min-width
screen max-width
screen min-width max-width
screen min-device-pixel-ratio
screen max-device-pixel-ratio
screen device-pixel-ratio
screen min-dpr max-dpr
screen pointer *
screen hover *
screen min-height
screen max-height
screen min-height max-height
screen min-device-width
screen max-device-width
screen device-width
screen min-dw max-dw
all min-width
all max-width
all min-width max-width
all min-device-pixel-ratio
all max-device-pixel-ratio
all device pixel ratio
screen min-dpr max-dpr
all pointer *
all hover *
```

```
all min-height
all max-height
all min-height max-height
screen min-device-height
screen max-device-height
screen device-height
screen min-dh max-dh
screen orientation
screen min-aspect-ratio
screen max-aspect-ratio
screen aspect-ratio
screen min-device-aspect-ratio
screen max-device-aspect-ratio
screen device-aspect-ratio
screen min-resolution
screen max-resolution
screen min-reso max-reso
screen resolution
all min-device-width
all max-device-width
all device-width
all min-dw max-dw
all min-device-height
all max-device-height
all min-dh max-dh
all orientation
all aspect-ratio
all device-aspect-ratio
all min-resolution
all max-resolution
all min-reso max-reso
all resolution
```

On top of this, the height/width/aspect ratio MQs are likely used often with print and TV media types, although I haven't used them myself.

For me, it's less of an "all or nothing" approach as it is a "we shouldn't limit ourselves with what is popular now" approach. While I agree the solution needs to be trivial to use, I disagree with the thought (and for that matter, the practice of) discrete mixins for each different possible media query when with a single one, we can not only cover the 90% of common use cases simply (and for that matter, spitballing roughly 70-80% of all written media queries with the defaults), but 100% of future media queries. The point of following mental model is well taken, but I feel as if the people who are going to be using Compass's RWD functionality are going to know what the term "breakpoint" is as it is a very common term in RWD and not need the hand holding of  something like "when-wider-than". As for being able to handle min/max in one mixin, with something as spelled out "when-wider-than" would be a min-width mixin, with your $until adding a max-width, but if we wanted the opposite, we
'd need "when-smaller-than", which is now two mixins for dam close to the same functionality. Then you have other issues, like DPR for -webkit and -moz both take a decimal, but -o takes an actual ratio, so the question becomes which do you use, as they're all vendor specific.

As for setting a variable with all breakpoints, that's not really something you can just write for best practice RWD, it's gleaned by your content, and some things will break at different points than others, say the navigation may break before the sidebar does. How would that (common) scenario be expressed? Or, for that matter, having a single variable with just breakpoint numbers, how would you go about using them? You and I understand nth(), but should we expect nth() to be an integral part of how our end users need to use breakpoints? Isn't a string name or variable name easier?

As for nesting in Sass, if I wanted to write the following media query:

``` css
/* MQ including single size images */
@media screen and (min-width: 200px) and (max-widht: 400px) and (-webkit-max-device-pixel-ratio: 1)
```

I'd need to write the following with nesting:

``` scss
@include when-wider-than(200px, 400px) {
  @include when-pixel-ratio(1, 'max') {
  }
}
```

I feel that that's a much worse syntax than what the breakpoint solution affords:

``` scss
$bkp1: 200px 400px, 1 device-pixel-ratio;

@include breakpoint($bkp1){}
```

The second is DRY while the first would be WET as you would need to nest that DPR query each time, or you could wrap it yourself in a mixin, adding more abstracting away. For sure, there's more work to be done to make Breakpoint have as seamless of a DX as possible, but I believe the potential for Breakpoint is much larger than the MQ specific mixins do, doubly so for the MQ specific approaches based on devices.

chriseppstein:
@Snugug I've already stated my objections to that API very clearly. It doesn't guide. It's not as readable as a normal MQ. It makes it harder to learn this module. I think Sass does a very good job at providing a minimal abstraction on top of @media and I am quite against introducing a full-fledged abstraction of @media in Compass.

Compass should augment what Sass provides with best practices and cross-browser compatibility. I'm ok with being opinionated in this respect. Maybe the best practices of RWD have not yet solidified. If that is the case, we shouldn't add a module to Compass, but I get the feeling that the RWD community _has_ developed best practices. Let's codify them.

anthonyshort:
The query syntax sounds like a perfect candidate for an extension but it's too abstract and will make the module difficult to learn. I don't think it fits in well with the goals of Compass. Again, with the example you gave I would just write the standard media query in a variable and use  @media. It gives the same result without being abstract. 

All of those media queries listed really aren't the most common. If you're using all of those then you are probably a special case that I don't think could be accounted for. What I think we need is a way to make the most common goals easier while not attempting to replace media queries all together.

What most people need are helpers for widths and heights. You can the dimensions of the page and content can change.

A single mixin (or 2) could do this. It would allow for min and max. This could possibly satisfy 90% of cases. Use variables and a mixin to simplify setting widths. This, along with nested and constructed media queries, could be enough. 

Media queries themselves aren't difficult especially with Sass.

Snugug:
Thank you, been doing this from my phone.

codingdesigner:
I found myself looking up the query syntax as I started each project. (Now, do I write screen first? Do I quote the features?) It's a long-form syntax and that's at least a small point of friction.

As for managing sets, when I work on an rwd project I end up with many points where something changes. The overall layout will change a few times, but there are many more times where smaller adjustments are made. For example at a certain point in my current project I change the positioning of the logo. If only that one thing happens at that point then there's no need to "manage" it, but I also want to change the positioning and font-size of the tag line, and the background of the header. Each of these items is in a different place in my code. At this point I want to name my breakpoint, so that I can be clear that these things happen at the same time and for the same reason. Over time these use cases build up.

```
$bp-headline-change: 53em

body.front
  #name-and-slogan
    +breakpoint($bp-headline-change)
      max-width: 33em

#preheader
  +breakpoint($bp-headline-change)
    position: relative
    .announcement
      position: absolute
      right: 0.5em
      top: 61px

.not-front
  #name-and-slogan
    +breakpoint($bp-headline-change)
      img.logo
        width: 50%
      .tag
        font-size: 1.25em
        margin-bottom: 0
```

I think the most common expectation is that devs will make breakpoints a few times when the layout changes. I think a more subtle approach is developing where less grand changes are happening more often. Change the leading or baseline based on the measure, tighten a little padding here, loosen a little there. Giving those breakpoint events a name makes it much easier to stay on top of these, and having a shorthand for that helps as well.

The `when-wider-than` syntax in intriguing. As I use Breakpoint I've come to a similar practice, but as a variable naming convention: `$bp-not-handhelds: $bp-handheld-width + 1`. I like how it guides you thru the logic of what will happen.

If we were to go in that direction with Compass then you could rewrite my example something like this: 

```
$compass-headline-change: 53em

.something
  +when-wider-than($compass-headline-change)
    color: red
```

That would cover most mainstream uses, but when it came to giving a name to compound queries you'd have to create a custom mixin. 

```
=rotate-phone
  +when-narrower-than(480px)
    +when-oriented(portrait)
      @content
```

That's pretty reasonable I think. It solves most of the issues for most of the people. I still think there's room for something more robust, but if you need that you probably also understand extensions. 

I do have an issue with the proposed naming. In practice, I might expect `+when-wider-than(320px)` to compile to `@media screen and (min-width: 321px)`. Perhaps using the feature name in the mixin would be more intuitive? 
- `when-min-width($width, $until: false)`
- `when-max-width($width: $from: false)`

chriseppstein:
Closing.

PR #866, opened 2012-04-28
add a sprite-image-position() helper function

kwliou:
This is would be really useful for for automatic sprite handling. I use Compass spriting in many group projects and often other people don't want to deal with advanced spriting.

I don't insert elements just for icons so for an icon that's supposed to be on the very right side of a block element, I would normally do something like:

```
$map-sprite-position: 100%;
...
.map-sprite {
  @include sprite-background-position($map-sprites, sprite, 100%);
}
```

Using this helper method, I can then write something like:

```
@mixin adjust-sprite-positions($map) {
  @each $sprite in sprite-names($map) {
    $position: sprite-image-position($map, $sprite);
    @if $position != 0 and unit($position) == "%" {
      .#{sprite-map-name($map)-$sprite} {
        @include sprite-background-position($map, $sprite, $position);
      }
    }
  }
}
```

### DIFF
--- a/lib/compass/sass_extensions/functions/sprites.rb
+++ b/lib/compass/sass_extensions/functions/sprites.rb
@@ -174,6 +174,23 @@ module Compass::SassExtensions::Functions::Sprites
     raise Sass::SyntaxError, %Q(The sprite-image() function has been replaced by sprite(). See http://compass-style.org/help/tutorials/spriting/ for more information.)
   end
 
+  # Returns the position of the sprite in the sprite map specified by $<map>-<sprite>-position
+  # This can be useful for automatically adjusting the background-position property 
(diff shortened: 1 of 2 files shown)


scottdavis:
im ok with this as long as there is a test case created

kwliou:
Ok I wrote a simple test

scottdavis:
@chriseppstein im ok with this if you want to merge it for the next rc

chriseppstein:
@scottdavis you have commit :P merge it if it's low risk.

chriseppstein:
Sorry, this merge introduced syntax errors that once fixed allowed the tests to fail. I need to release stable, so I reverted this. Please submit a patch against master for the 0.13 release.

PR #873, opened 2012-05-03
Fixing positioning issue

vbichkovsky:
When there are some repeat-x images in the sprite and some other images have position set to 100%, there is a gap between sprite's right side and images with position = 100% (because width-fix was applied after they had been positioned).
I tried to figure out how to write a test for such case and I must say that your tests are difficult to comprehend ;)


chriseppstein:
@scottdavis can you review and merge when convenient?

lib/compass/sass_extensions/sprites/sprite_methods.rb
```diff
@@ -1,3 +1,5 @@
+require 'rational'
```
scottdavis, line 1: Do we need this require?

vbichkovsky, line 1: Nope, it just happened that my commit contained these changes and I created a pull request from it. You only need to change layout_methods.rb

scottdavis, line 1: manually changing something kinda makes using pull requests pointless =p

chriseppstein:
@scottdavis what's the status here? Are you holding this up so that a single require can be removed? >_<

scottdavis:
No I'm holding it up because I'm busy I'll get to it this week I'm going to cherry-pick it into master

scottdavis:
merged 5e42b8f57d885fa603c07908a272003c020d3d9b

PR #877, opened 2012-05-08
Adds text-shadow fallback for browsers that do no support blur

nearestnabors:
(See also IE' s text-shadow implementation). 

Note that this does not add a fallback for multiple text-shadows, as those do not use variables. 

Addresses issue 870.

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_text-shadow.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_text-shadow.scss
@@ -43,6 +43,7 @@ $default-text-shadow-blur:     1px  !default;
   @if $color == none {
     text-shadow: none;
   } @else {
+    text-shadow: $color $hoff $voff; // for browsers without spread distance (like IE)
     text-shadow: $color $hoff $voff $blur;
   }
 }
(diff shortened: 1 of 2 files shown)


thefort:
Seems like, if there are any shadows with 4 arguments, truncate those to three and emit that text-shadow first. I guess we'd only do this if `$experimental-support-for-microsoft` is true.

Can you think of any reason why we wouldn't do this?

/cc @ericam @imathis @anthonyshort

frameworks/compass/stylesheets/compass/css3/_text-shadow.scss
```diff
@@ -43,6 +43,7 @@ $default-text-shadow-blur:     1px  !default;
   @if $color == none {
     text-shadow: none;
   } @else {
+    text-shadow: $color $hoff $voff; // for browsers without spread distance (like IE)
```
mirisuzanne, line 46: `spread` and `blur` are different things. I think everyone supports blur, am I wrong? The issue with IE, as I understood, is that they are adding spread. So we would need to output one more variable than we have been, not one less.

nearestnabors, line 46: You're right. It's a fourth value. I have hideously embarrassed myself by getting it backwards.

mirisuzanne, line 46: I'm sure you'll recover somehow. :)

scottdavis:
whats the status on this?

nearestnabors:
Currently at conference. Solution possibly beyond my Ruby abilities.

scottdavis:
ping @ericam

chriseppstein:
I did this in b726f822b7fb9c3159a718f0cdde09dcf5842d2e

Issue #836, opened 2012-04-06
Nested Sprites docs typo


ghost:
Hello, 

http://compass-style.org/help/tutorials/spriting/

```
<p>Example:</p>

<pre><code>  @import "themes/orange/*.png";
    @include all-orange-sprite;
</code></pre>
```

I believe the last `@include all-orange-sprite` in the example statement needs an `s` added to sprites. 

Correction: `@include all-orange-sprites;`

I was getting compile errors before I noticed the "unpulaized" sprite. 

```
Syntax error: Undefined mixin 'all-foo-sprite'.
```

PR #888, opened 2012-05-10
Fix documentation issue #836

jphpsf:
I believe issue #836 is valid. In some of my SCSS code, I actually use this syntax:

```
$notification-spacing: 75px;
@import "notification/*.png";
@include all-notification-sprites;
```

I also verified, if I remove the s, get the following error:

```
    error sass/notification.scss (Line 3: Undefined mixin 'all-notification-sprite'.)
```


chriseppstein:
Thanks for the patch!

PR #893, opened 2012-05-16
Allow for Inline Images in replace-text mixin

Snugug:
Recently came across the want to use the replace-text mixin with an inline image, and found I couldn't, so here's a small little patch to allow that!


chriseppstein:
Thanks for the patch. I ended up not using it because I decided the API that compass was presenting was needlessly restrictive. You can now just do this: `@include replace-text(inline-image("foo.png"))` However, the `replace-text-with-dimensions` mixin does now accept an `$inline` argument so that it can still read the dimensions.

Snugug:
Great to hear, thanks!

PR #894, opened 2012-05-17
Fix finding bundler

mpapis:
it's related to this SO answer: http://stackoverflow.com/questions/10610254/cant-install-compass-via-rvm/10611366#10611366

I was talking to @wycats and there are two possible solutions:

1) detect if Gemfile is in effect - this pull request:

```
if defined?(Bundler)
  require 'bundler/shared_helpers'
  Bundler.require :assets if Bundler::SharedHelpers.in_bundle?
end
```

2) detect if `Bundlr.setup` was called:

```
if defined?(Bundler) && Bundler.instance_variable_defined?("@setup")
  Bundler.require :assets
end
```

### DIFF
--- a/bin/compass
+++ b/bin/compass
@@ -22,7 +22,8 @@ fallback_load_path(File.join(File.dirname(__FILE__), '..', 'lib')) do
 end
 
 if defined?(Bundler)
-  Bundler.require :assets
+  require 'bundler/shared_helpers'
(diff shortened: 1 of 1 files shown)


chriseppstein:
merged as a368a574137ce4d89bd85a99639c7473e3971456

chriseppstein:
Thanks for the patch.

Issue #775, opened 2012-03-13
Add rem support with pixel fallbacks to vertical-rhythm module


ry5n:
Currently the vertical-rhythm module supports setting the unit of font-size to rems, but doesn't provide pixel fallback declarations for older browsers and [has some other issues](https://github.com/chriseppstein/compass/pull/766#issuecomment-4450905). More complete support for rems, including pixel fallbacks, would be an awesome addition to this module. See [#766](https://github.com/chriseppstein/compass/pull/766). [#771](https://github.com/chriseppstein/compass/issues/771) is also relevant.

Additional rationale:
- Would help authors by removing the need to ever pass the $font-size of the current context. This can be a pain not only because the context can't be computed but because it makes the API more complex and harder to learn.
- Would improve maintainability of projects as it removes hard-coded $font-size contexts.

Considerations:
- Using rems means doing extra compilation math and also doubles up on style rules for legacy support (not as much bloat as vendor prefixes mind you)
- Using rems requires setting the base font size on the root element rather than on the body. There may be issues with older browsers in doing so.

JohnAlbin:
Its definitely telling that we were both working on this at the same time. :-)

chriseppstein:
I think rem is a great unit for developer authoring, but I don't see a need to change the output unit to rem. I would support any changes needed to make it simple for authors to pass rem-based values to the compass mixins but when using absolute units -- let's keep the px output. Please let me know if I'm missing something.

Also, seems like this could be done as non-breaking addition to stable. If not, it'll need to go onto master for 0.13.

ry5n:
@chriseppstein I'm not 100% sure what you're suggesting and I'd like to understand. Are you suggesting we allow rem inputs (with rem outputs) but no pixel fallbacks? I think that that's already supported (except for +establish-baseline operating on the body rather than root).

Maybe I can restate the rationale for this better by laying bare my thought process:
1. Sizing fonts in px is bad (received wisdom; mea culpa – is the only reason because of IE6's font-resize issues?)
2. Sizing fonts in ems is tricky (context has to be tracked by the author)
3. Rems solve everything :)
4. Except they aren't supported by older browsers :(
5. But we can use rems today and offer some support to older browsers by writing pixel fallback values (and yes oldIE will still
   have issues): http://blog.typekit.com/2011/11/09/type-study-sizing-the-legible-letter/
6. If this is at all a sensible approach***, it would be great to have as an option in vertical-rhythm.

[...]

PR #896, opened 2012-05-20
Add rem output with pixel fallbacks and other improvements to vertical rhythm partial

ry5n:
I've implemented support for feature #775 in this branch. However, there are other changes I'm proposing here, a few of which are not backwards compatible. @JohnAlbin and @ericam, I'd love to get your feedback in particular.

Full explanation of changes is below and the diff with master is [here](https://github.com/chriseppstein/compass/pull/896/files). Oh, test fixtures are included too and run as expected.

---
## Architectural changes

`$font-unit` is replaced with a new `$rhythm-unit` variable which accepts a string representing the desired unit for rhythm output: 'em', 'rem' or 'px' (default remains 'em'). Percentage units are no longer allowed because they make calculating margins and padding impractically complex and actually break the rhythm-border mixins completely ([percentage borders are invalid css](http://lists.w3.org/Archives/Public/www-style/2012Jan/0084.html)).

When the output unit is 'rem', by default a fallback declaration in pixels is output for older browsers. This naturally increases output file size, and can be turned off by setting a new `$rem-with-px-fallback` flag to false.

`$relative-font-sizing` is now derived from `$rhythm-unit` and marked private. Previously both `$font-unit` and `$relative-font-sizing` were configurable, with the value of `$font-unit` depending by default on the value of `$relative-font-sizing`. This branch inverts that dependency and removes `$relative-font-sizing` as a concern for authors. Now there is only one thing to set and it directly and accurately expresses what it is.

Internally, almost all unit conversions are now handled by a `convert-length` function. The function is fully abstracted and handles converting between any css lengths/percentages. There is also a new `rem` mixin that handles the logic and output of pixel fallback declarations for rems. Both of these live in a new `units` partial, which vertical-rhythm now imports. `$browser-default-font-size` has been moved here as well.

Convert-length gracefully handles input and has warnings for things that will break, which has allowed me to remove many of the warnings of this type: `@warn "$relative-font-sizing is false but a relative font size was passed to…"`.

Also thanks to convert-length, it's now possible to be more liberal in the units accepted by mixins. The main use case is the `adjust-font-size-to` mixin, which now supports passing the `$to-size` parameter in any valid css length or as a percentage.

However, `$base-font-size` and `$base-line-height` are still assumed to be in pixels. In fact, this is now enforced with a warning, since a few things still break if they aren't. They could be made to support the full range of font units (and represented internally as px), but given they are specified only once, I'm not inclined to do it.
## Other changes

I've always had trouble using the rhythm mixin because it handles margins and padding all at once, and has less-than-ideal parameter defaults. Its signature has been optimized for the most common use case (symmetrical rhythm spacing above and below an element):

Old:

``` scss
rhythm($leader: 0, $padding-leader: 0, $padding-trailer: 0, $trailer: 0, $font-size: $base-font-size)
```

New:

``` scss
rhythm($leader: 1, $padding-leader: 0, $padding-trailer: $padding-leader, $trailer: $leader, $font-size: $base-font-size)
```

I've added two new mixins in the same family: `rhythm-margins` and `rhythm-padding`, which apply symmetrical whitespace as you'd expect.

Mixin `$apply-side-rhythm-border` now supports keyword 'all' for its `$side` parameter, which produces output using the border shorthand rather than four sets of identical declarations.

One more thing: the rhythm-border mixins now support border colors. Authors can set $default-rhythm-border-style to eg. `solid #aaa` instead of the default `solid` and get their desired border color for free. Authors can also pass a similar two-value list individually to any of the rhythm-border mixins.


frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
```diff
@@ -24,62 +29,107 @@ $round-to-nearest-half-line: false !default;
...
 // Show a background image that can be used to debug your alignments.
+// As this is a development feature, this mixin never outputs pixel fallbacks
+// for rem output.
 // Include the $img argument if you would rather use your own image than the
 // Compass default gradient image.
 @mixin debug-vertical-alignment($img: false) {
   @if $img {
     background: image-url($img);
   } @else {
-    @include baseline-grid-background($base-rhythm-unit);
```
ry5n: $base-rhythm-unit was only ever used in this one spot, and has been removed.

mirisuzanne: I've been using `$base-rhythm-unit` as a useful part of the public API. Is there a reason not to keep it available?

mirisuzanne:
Very cool! Most of this sounds good to me. I'll have to play with it some to see how it works exactly.

frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
ry5n: There's no reason to get rid of it if it's useful, just that AFAIK it's identical to a call to rhythm().

ry5n:
Awesome! Let me know how it goes :)

frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
mirisuzanne: You're right. Good riddance. :)

ry5n: :) Assuming this gets merged, it should be an easy find-and-replace.

scottdavis:
+1 and another +1 for tests ... only problem is its not a clean merge =(

ry5n:
@scottdavis Yay. Also: rats. How can I fix the merge problem?

Sort of off-topic: once the API is stable again I'd like to write a demo for the docs. :)

scottdavis:
need to rebase

ry5n:
Rebased master onto this branch, removed from github and re-published the rebased version (same branch name). Merge appears to be clean now, at least locally. BTW if I totally did that wrong be as blunt as you like.

scottdavis:
omg a correct rebase i am excited!

ry5n:
WOOO

scottdavis:
whats the status on this /ping @ericam @chriseppstein

mirisuzanne:
I won't have time to look at it for another week. Happy to check it out (and merge it) then, unless someone beats me too it.

ry5n:
@scottdavis Thanks for checking in. @ericam I’d rather this receive a through review before merging, so take the time you need. Cheers :)

ry5n:
@ericam, @scottdavis, et. al. Anyone had a chance to review this? Anything I can do to help?

mirisuzanne:
From what I understand of it, this looks pretty good to me. There's a lot here, so it'd be nice to see an overview of all the changes. It doesn't break the tests, and everything I've tried worked as expected, so a few quick questions and I'm ready for it.
1. Isn't all the unit conversion stuff part of Sass? Was it necessary to rebuild it here? Or does this cover more territory in some way? If so, should it move into Sass?
2. With this much control over units, is there a reason we continue to force `$base-font-size` and `$base-line-height` into px values? Should we accept any value (px, em, %) there instead?

ry5n:
\1. 
    AFAIK Sass handles unit conversion between absolute length values only (like `in` to `px`, `cm` to `pt`) using a conversion table. However, it can't do that with relative units, including em, rem and percent. To help with this patch I needed to deal with mostly relative units so I wrote something to help with that.

Usage of the `convert-length` function in this patch is relatively straightforward because the input and output units are all font-related. For other potential conversions things get much trickier because the unit being converted _from_ might be relative to one metric while the unit being converted _to_ may be relative to a different one. For example, say I want to convert `rem` to `%`: as a stylesheet author, I not only have to know the font-size of the root element ($base-font-size) but also the context length to which % refers. Those things have to be supplied as extra arguments, for a total of up to four: input length, output unit, input context and output context. Add to that the fact that percentages in CSS may be relative to different things depending on the property on which they are applied…

So I believe this is really only useful for frameworks. _Could_ be really useful, but it's the part of this patch I'm most nervous about. I want it to be a net gain for the framework.

\2.
We could do that. I even almost did it. But those values would have to be converted internally, and then the user-supplied values (used elsewhere by the author and by other frameworks) would be harder to work with. A percentage-based $base-font-size might have unexpected (though correct) results if naïvely applied as padding or margin, for instance. I'd be curious to know what you think, though.

mirisuzanne:
That makes sense to me.

I'd vote for #2 happening (with internal conversion) if it's possible, but I don't think it's necessary as part of this commit. Thumbs up from me. If no one else jumps in here with a veto or more questions, I'm ready to pull this in.

ry5n:
OK, sounds good then. If there are no other objections, feel free to merge. Once that's done, I can add the internal conversion. I will write a vertical rhythm demo for the docs site as well; I think a lot of folks remain confused about how to use this partial.

frameworks/compass/stylesheets/compass/typography/_units.scss
```diff
@@ -0,0 +1,120 @@
...
+	@if unit($to-context) != 'px' { @warn "Parameter $to-context must resolve to a value in pixel units."; }
+
+	// Fixed ratios
+	// https://developer.mozilla.org/en/CSS/length
+	// http://dev.w3.org/csswg/css3-values/#absolute-lengths
+	$px-per-in: 96px / 1in;
+	$px-per-mm: 96px / 25.4mm;
+	$px-per-cm: 96px / 2.54cm;
+	$px-per-pt: 4px / 3pt;
+	$px-per-pc: 16px / 1pc;
```
chriseppstein: This stuff should be baked into the Sass compiler, I have opened a new issue for that: https://github.com/nex3/sass/issues/471

frameworks/compass/stylesheets/compass/typography/_units.scss
```diff
@@ -0,0 +1,120 @@
...
+//
+@mixin rem($property, $values, $use-px-fallback: $rem-with-px-fallback) {
+	// Create a couple of empty lists as output buffers.
+	$px-values: ();
+	$rem-values: ();
+
+	// Ensure $values is a list.
+	@if type-of($values) != 'list' {
+		$values: join((), $values);
+	}
```
chriseppstein: Sass will treat a single value as a list with one element in an `@each` loop.

``` scss
@each $foo in 2px { div {width: $foo; }}
```

outputs:

``` css
div { width: 2px; }
```

frameworks/compass/stylesheets/compass/typography/_units.scss
```diff
@@ -0,0 +1,120 @@
...
+
+	// Use pixel fallback for browsers that don't understand rem units.
+	@if $use-px-fallback {
+		#{$property}: $px-values;
+	}
+
+	// Use rem values for everyone else (overrides pixel values).
+	#{$property}: $rem-values;
+}
```
chriseppstein, line 152: I've come to really dislike any code that operates on Sass lists. It's really quite unpleasant to write or read. So I opened a new issue for sass to add anonymous functions: https://github.com/nex3/sass/issues/472

frameworks/compass/stylesheets/compass/typography/_units.scss
chriseppstein: I made a patch for sass to address the `96px` per `1in` thing: https://github.com/nex3/sass/pull/474

chriseppstein: and it was merged :)

ry5n: Ah, thanks Chris!

frameworks/compass/stylesheets/compass/typography/_units.scss
ry5n: Wow, glad to be wrong about this! I'll refactor this and update.

frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
```diff
@@ -24,38 +29,75 @@ $round-to-nearest-half-line: false !default;
...
+      @else {
+        $output: join($output, $value);
+      }
+    }
+    #{$property}: $output;
+  }
 }
 
 // Establishes a font baseline for the given font-size.
 @mixin establish-baseline($font-size: $base-font-size) {
```
ry5n, line 96: I just discovered a critical bug! For rem units, if $base-font-size is not 16px, the line-height will not be correctly computed in establish-baseline, because on the html element 1rem is always 16px (the initial value). Right now 1rem is treated as if it were $base-font-size. I need to solve this before this can be pulled in.

ry5n, line 96: This problem appears to be restricted to Webkit, and may be a bug. [The spec](http://www.w3.org/TR/css3-values/#rem-unit) says that 1rem is equal to the computed value of `font-size` on the root element, except for `font-size` on the root itself (in which case 1rem is equal to the initial value). However, in Webkit it appears that `line-height` on the root element also uses the initial value rather than the computed value. One possible work-around could be to set line-height on the body while setting font-size on html. Another could be to set line-height in ems in `establish-baseline` only, even when `font-unit` is `rem`. I need to investigate further, and also file a Webkit bug.

ry5n, line 96: Just to clarify, this bug has to do with _computed_ values in Webkit. The values output by Sass were always correct (and seem to be to-spec), which is why tests didn't pick them up. What happens is that if `html { font-size: 112.5%; line-height: 1.4rem }` the computed value of `line-height` in Webkit is 1.4 \* **16px** = 22px when it should be 25px.

JohnAlbin, line 96: > One possible work-around could be to set line-height on the body while setting font-size on html.

That's fugly.

> Another could be to set line-height in ems in establish-baseline only, even when font-unit is rem.

Much better idea! :-D

mirisuzanne, line 96: I agree. There is no need for "root ems" when you are at the root. Any ems on the html tag are root ems!

ry5n:
I'm swamped with work work at the moment and won't have a chance to update this pull request until the first week of September, probably. When I do, my plan is to:

a) fix the Webkit establish-baseline issue by setting line-height on html in ems for all relative rhythm-units. Has been working for me so far in testing;

b) ditch the unnecessary checking for lists in @ each;

c) switch to using Sass's built-conversions for units when possible.

ry5n:
One other thing: @chriseppstein is there a good reason to be rounding px here? IIRC I added the rounding code to make tests pass, but do we mind decimal pixels in stylesheets?

silvenon:
While you're adding a lot of sexy features, it would be cool if you could also add a mixin which accepts a block and temporarily sets the `$base-font-size` and `$base-line-height` for that block. Something like this:

``` scss
@mixin with-rhythm-settings($font-size, $line-height) {
  $tmp-font-size: $base-font-size;
  $tmp-line-height: $base-line-height;

  $base-font-size: $font-size;
  $base-line-height: $line-height;

  @content;

  $base-font-size: $tmp-font-size;
  $base-line-height: $tmp-line-height;
}
```

Pretty useful when using a lot of rhythm functions and mixins for font sizes other than `$base-font-size`.

ry5n:
@silvenon Could you write a little about your specific use case? Example?

silvenon:
I realized setting a different line height doesn't make sense because it would break the rhythm. So, something like this:

``` scss
@mixin with-font-size($to-size, $lines: lines-for-font-size($to-size), $from-size: $base-font-size) {
  @include adjust-font-size-to($to-size, $lines, $from-size);

  $tmp: $base-font-size;
  $base-font-size: $to-size;

  @content;

  $base-font-size: $tmp;
}
```

You can use it when dealing with this kind of beasts:

``` scss
article {
  $font-size: 14px;
  @include adjust-font-size-to($font-size);
  @include trailer(1, $font-size);
  @include h-borders(1px, .5, $font-size);
  padding-left: rhythm(.5, $font-size);
  p  { @include trailer(.5, $font-size); }
  ol { @include trailer(1, $font-size); }
  li { @include trailer(.25, $font-size); }
}
```

Instead you could do:

``` scss
article {
  @include with-font-size(14px) {
    @include trailer;
    @include h-borders(1px, .5);
    padding-left: rhythm(.5);
    p  { @include trailer(.5); }
    ol { @include trailer; }
    li { @include trailer(.25); }
  }
}
```

It's pretty useful when you decide to change the `$font-size` back to `$base-font-size`. Then you can delete only two lines, rather than going through each line to delete that `$font-size`.

ry5n:
@silvenon That's actually a really clever way to deal with the compounding-em problems that motivated me to make the main changes featured here. Except it's handled in pure Sass. I can't see us _not_ having this. :)

silvenon:
Yay :) Can't wait for the pull :D

silvenon:
Also, it would be useful to have a mixins that remove rhythm borders, rather than having to do things like:

``` scss
@include apply-side-rhythm-border(left, 0px, 0);
```

Or maybe I've been doing it the wrong way the whole time: )

Also also, `v(ertical)-borders` mixin would be nice, and generally a more encouraging support for side borders. I know side spacing is not really in the vertical rhythm spirit, but often I want the same spacing on the sides too, so it's better to have those mixins than having to rape that `rhythm()` function all the time.

mirisuzanne:
I would use v-borders as well. There are several other bit of touch-up I would like to do, but I don't think it all needs to happen in the span of this one pull. I'd like to get these basics in asap and then look at what else might be needed.

ry5n:
I'm still finishing up the extra work-work projects that have kept me occupied but will have time at the end of this week to  finish this up. I'll work on follow-up soon after that.

ry5n:
- Fix the Webkit `establish-baseline` issue (see above) by setting line-height on <html> in `em`s for all relative rhythm-units.
- Remove unnecessary type-checking for Sass lists.
- Switch to using Sass's built-in (absolute) unit conversions when possible.

This passes tests but I want to run it through a few extra paces with some real-world stylesheets.

ry5n:
I've done some additional testing with these changes and this should be good to go. I'd like @chriseppstein and/or @ericam to take a look and see if there’s anything I missed, otherwise let’s get this in!

torkiljohnsen:
A blog post on this topic surfaced today, outlining a better fallback
alternative than pure pixels.

http://seesparkbox.com/foundry/scss_rem_mixin_now_with_a_better_fallback

ry5n:
@torkiljohnsen That's an interesting approach. It's certainly not a silver bullet, since keyword values for font-size may not work well for all designs. There would be extra overhead, either in the mixin or cognitive, to determine which keyword is closest to the desired font size. But the whole reason for using rems over ems is to remove the cognitive overhead of minding the current font-size. On balance, I'm tempted to say that if page zoom in IE7/8 isn't producing accessible results for a particular project, or if IE6 support is needed, the author should not use an experimental font-sizing approach like rems. Bite the bullet and use ems. Come to think of it, perhaps this mixin should @warn if $font -unit is 'rem' and legacy IE6 support is true.

ry5n:
@chriseppstein @ericam @scottdavis Any chance we can get a final review/approval on this? It’s finally ready.

mirisuzanne:
I'm a bit swamped, but will try to get to it sometime this week. Excited!

ry5n:
@ericam Thanks! Me too!

mirisuzanne:
Great work! Thanks!

ry5n:
Oh wow! Yay! Thanks Eric. Now it’s time for me to write a full example for the docs, and see about some follow-ups :)

mirisuzanne:
\o/ - And it's time for me to move all my projects over to rem. :)

JohnAlbin:
Yay! Nice work, everyone! (Especially ry5n!)

silvenon:
I can't believe I can stop using ems now, that was so annoying. Thanks a lot :)

krisbulman:
Hope this makes the upcoming pre-release!

scottdavis:
Its already merged so it will

ry5n:
I want to add an example to the docs, but for this to work I need a full HTML tree _within_ the example for the mixins to work on. I'm thinking this means an iframe (?). But I am not familiar with how the docs site is built from its source and this may be an edge case. Can someone familiar with the examples part of the Compass docs site point me in the right direction?

ry5n:
Once the docs are updated, I will look at follow-ups, including
- Adding a with-font-size() mixin
- Consider adding additional border mixins to help with removing rhythm-borders and applying left and right borders.
- Consider adding a warning for rem units when IE6 support is true (since it does not have page zoom)

Let me know if I missed anything in that list.

silvenon:
Concerning the docs, I got you covered. I suggest waiting until #1070 gets pulled, then you'll find all you need to compile the site locally, generate examples and other stuff in doc-src/README.markdown.

pyronaur:
I want it all, I want it all, and I want it now. 
Is it in Compass 0.13 alpha gem ?

silvenon:
Not yet, for now use:

``` rb
gem "compass", github: "chriseppstein/compass", branch: "master"
```

You can find out how it works from the source code, there's no documentation yet (except for the comments).

ry5n:
Sorry to slack on docs; not what I aspire to. I'm setting aside time this weekend to maker sure things are up-to-date on master, and start on a demo.

torkiljohnsen:
Me wants! When can we expect the next Compass release btw? The list of open issues on 0.13 looks long. Perhaps some should be moved to a future release so this awesome feature can be released? :) :+1:

krisbulman:
@torkiljohnsen next alpha pre-release ( #1068 ), has a milestone with 3 issues open: https://github.com/chriseppstein/compass/issues?milestone=11&state=open  I think that may be the best bet for a release in the near future if you're looking for issues to help out with.

rctneil:
Is this in Compass 0.13.alpha.0?   I tried using the vertical rhythm module but it still outputs ems? I have the following variables set:

$base-font-size: 16px
$base-line-height: 24px
$rhythm-unit: 'rem'
$rem-with-px-fallback: true

I get em output and I need rem with px fallbacks on output.

Also, is the rem() mixin now in this version of Compass?

silvenon:
No, I think it's in 0.13.alpha.2. Current version of Compass is 0.13.alpha.4 so you can just update to that one.

timkelty:
What do people think about changing the `$to-unit` param in `convert-length` to have a default of $rhythm-unit?

If you were setting your `$rhythm-unit` to `rem`, you're basically always going to be converting to that, right?

ry5n:
@timkelty Thanks for your interest! `convert-length` is designed as a generic conversion function. For that reason, it can’t assume a particular use case. It’s also awkward to use in day-to-day Sass – it’s mostly useful for frameworks or extensions. I don’t want to tie it tightly to vertical rhythm.

Still, it might help to understand your use case to better explore what you’re running into.

timkelty:
@ry5n Gotcha, makes sense to keep it generic.

I was just using it for setting lots of rem variables (for font sizes and Susy grids), so I was using `$rhythm-unit` and `$base-font-size`, but not necessarily using Compass Vertical Rhythm.

I guess for the rest I can just use the `rem` mixin, so it's all good.

Thanks!

ghost:
@rctneil mentioned a rem() mixin a few comments up. What is this? Is there a mixin in compass now for converting to rems with px fallbacks? If so, where can I find out more?

rctneil:
There is a rem mixin in the latest github build. It works like this: +rem(padding-left, 10px)

Works really nicely.

ghost:
Cool. That looks a lot like one that ry5n did (https://github.com/ry5n/rem) that I'm already using - I'm guessing it's the same.

Any docs you can point me to?

ry5n:
@dixhuit It is the same. You’ve reminded me that the docs on my github repo are kindof lacking. Basically, I you want to set a property value in rem units with a fallback in pixels, you would write:

```
@include rem(font-size, 24px);
```

or

```
@include rem(font-size, 1.5rem);
```

starryeyez024:
I tried to update to the latest version of compass for this reason and ran into a problem with the version of Listen that compass required was incompatible with the version of Listen that grunt required. Has anyone successfully used the latest build with grunt?

rctneil:
I too have had this problem with another gem that has Listen as a dependency. I too would be grateful for a resolution of this.

ry5n:
I haven’t used Grunt to manage Sass compilation, so unfortunately I can’t help with the dependency conflict. It probably warrants a separate issue though. If you create one, you can reference it here for others to find.

starryeyez024:
Thank you rctneil but I believe there is already an open issue for this problem:
https://github.com/chriseppstein/compass/issues/1306

PR #901, opened 2012-05-24
It doesn't need two selectors for this hack. Less selectors, less bits. ...

kud:
...Plus it needs hasLayout (with zoom: 1) to work perfectly. And it's "stack" and not "box" for Firefox 1 and 2.


chriseppstein:
merged as 84e9a684b9697d728a37abb14cb0aae2c4d2a790

chriseppstein:
Thanks for the patch!

PR #902, opened 2012-05-24
Fix gradient-type arg

kud:
When

```
@include filter-gradient(rgba(255,255,255,0.2), rgba(255,255,255,0.7), 0);
```

it generates

`filter: progid:DXImageTransform.Microsoft.gradient(gradientType=1, startColorstr='#33FFFFFF', endColorstr='#B3FFFFFF');`

 which should be

`filter: progid:DXImageTransform.Microsoft.gradient(gradientType=0, startColorstr='#33FFFFFF', endColorstr='#B3FFFFFF');`


chriseppstein:
The API for the filter-gradient is not to accept `1` or `0` for the 3rd argument. it is to accept `vertical` or `horizontal`. I have updated the docs in 6a1ca52 to make this more clear.

PR #905, opened 2012-05-26
Retrieve image properties from filesystem-like objects

johnbintz:
I want to use a MongoDB `GridFileSystem` as a source of images for Compass, and need to be able to analyze the images in the filesystem for dimension information. This patch provides the configuration parameter `images_file_source`, which accepts a `Proc` or `lambda` that accepts the path of the file to be analyzed and returns an `IO`-like object that can be used in the `ImageProperties` analysis. With this setup, using a `GridFileSystem` source for image data can be done like this:

``` ruby
require 'forwardable'

class CompassGridIO
  extend Forwardable

  def initialize(io)
    @io = io
  end

  def_delegators :@io, :read, :close

  def readbyte
    @io.getc.bytes.first
  end
end

Compass.configuration.images_file_source = lambda { |file|
  file = file.gsub(%r{^[\./]+}, '') # got to rip off any leading path characters

  CompassGridIO.new(grid_file_system.open(file, 'r'))
}
```

Having a file in `grid_file_system` at, say, `file.jpg`, will be accessed correctly when using any of the image-size methods in Sass:

``` scss
#logo {
  width: image-width('file.jpg');
  height: image-height('file.jpg');
  background: image-url('file.jpg');
}
```

### DIFF
--- a/lib/compass/configuration.rb
+++ b/lib/compass/configuration.rb
@@ -40,7 +40,8 @@ module Compass
       :preferred_syntax,
       :disable_warnings,
       :sprite_engine,
-      :chunky_png_options
+      :chunky_png_options,
+      :images_file_source
     ].flatten
 
     ARRAY_ATTRIBUTES = [
--- a/lib/compass/sass_extensions/functions/image_size.rb
+++ b/lib/compass/sass_extensions/functions/image_size.rb
@@ -4,7 +4,7 @@ module Compass::SassExtensions::Functions::ImageSize
     width, _ = image_dimensions(image_file)
     Sass::Script::Number.new(width,["px"])
   end
-  
+
   # Returns the height of the image relative to the images directory
   def image_height(image_file)
     _, height = image_dimensions(image_file)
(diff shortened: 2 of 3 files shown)


scottdavis:
im kinda thinking this may make something in compass-rails easier also +1

chriseppstein:
This feels like a awful lot of complexity to introduce to handle such an uncommon scenario. there isn't some way to mount this as a real filesystem at the unix level? I wouldn't mind introducing an image search path concept.

Beyond that, I don't understand why the abstraction for file finding sits at the image level, it seems more fundamental and if we were to add it, it should work for any asset type.

@scottdavis what does this solve for compass-rails?

scottdavis:
Finding images outside the assets/images directory but I think I'm dumb and sass-rails does that

chriseppstein:
@scottdavis the image-url function does, but I don't think we have tied internal image path lookups to that in compass-rails.

johnbintz:
@chriseppstein Mounting MongoDB GridFS as a real filesystem doesn't work, and development on a FUSE plugin for MongoDB stopped a few years back. I would be fine with pushing that abstraction farther up in the process, but ideally, being able to hand an `IO`-like object to anything that reads/writes media files for analysis/spriting/whatever would be ideal, as opposed to hard-coding in the use of `File.open`.

johnbintz:
If this doesn't fly I can just create my own Sass functions to do this and load them after Compass loads, so it's not a huge deal.

chriseppstein:
I'd really prefer to not add this complexity. But I will accept refactorings that make it easier to monkey patch the code to provide your own implementation.

johnbintz:
It's all good. Thanks.

PR #930, opened 2012-06-15
Adjusting the multi-column mixins to match the spec.

aarongustafson:
The implementation of column-rule-color, column-rule-style, and column-rule-width was missing the critical "column-" prefix for the properties.


aarongustafson:
See: http://www.w3.org/TR/css3-multicol/

mirisuzanne:
Are there any tests here that need to be updated? Is this ready to merge?

PR #936, opened 2012-06-21
Update Heroku screenshot

max:
The Heroku status site is a little more sophisticated and up to date as far as Compass goes. All in all it's a better showcase than the add-ons site. Replaced the Heroku screenshot.


chriseppstein:
can you cherry-pick this to stable and send it one more time?

max:
Sure thing.

max:
@chriseppstein What about #938? I couldn't get the docs to compile without it.

chriseppstein:
@mschoening I think I have a fix for that on stable, but it didn't get merged to master. I'll sort it out -- will pull that change in if I need it.

PR #938, opened 2012-06-21
Fix hyphenation doc

max:
While working on #936 I had trouble getting the docs to compile. They were looking for the stylesheet `compass/css3/_hyphenate` instead of `compass/css3/_hyphenation`. I updated the docs to reference the correct stylesheet and moved the file to `hyphenation.haml`. In the process I also fixed a typo (`/s/hypens/hyphens`).


chriseppstein:
this was fixed when I merged stable to master. Sorry for inconvenience!

PR #961, opened 2012-07-11
Adding hyphens support for Opera and IE via prefix.

### LINKED ISSUE #960: Add hyphens support for IE and Opera
Hi,

please add IE and Opera support to the hyphens mixin. Both browsers, IE 10 and Opera, are supporting this attribute via prefix.

Whissi:
Closes #960


scottdavis:
Needs a test case

chriseppstein:
This is on master now. Thanks!

PR #962, opened 2012-07-16
blueprint form elements have the wrong width

rizkysyazuli:
form element width should compensate with it's padding and border. updated to include the 5px padding and 1px border.


scottdavis:
Blueprint has moved to https://github.com/compass/compass-blueprint

PR #966, opened 2012-07-18
Sprite archetype

beausmith:
When a group of sprites all have the same width and height there is no clean way to only output the common width and height with the main `.<name>-sprite` class. The `$<name>-sprite-archetype` variable offers the ability for a user to specify a sprite from which the user can set the base width and height of the sprite.

Room for improvement: auto detect a common size of sprites and optionally output width and height if all sprites have the same dimensions.

(Sorry, this branch doesn't pass tests, couldn't figure out how to determine the correct generated sprite map file name to place into the test)


chriseppstein:
As we discussed face to face, I think there's some good ideas and goals in here, but the "archetype" approach isn't quite what I'd like to see to simplify the output. Still want to try the other approach?

PR #975, opened 2012-07-27
Fix invalid method call for strip_trailing_seperator in configuration

dannymidnight:
Fix `strip_trailing_seperator` method call when adding custom configuration method.


chriseppstein:
Not sure what this fixes since it doesn't have a test case.

dannymidnight:
Righto, I'll get on that.

chriseppstein:
This was merged to master.

PR #970, opened 2012-07-19
First blush attempt at addressing hi-res sprites

aarongustafson:
I have added the .rb tweaks to my local Compass Gem (0.13 Alpha 0) and they seem to be working well. Here’s how the changes play out:

``` scss
.slidecontrols {
    a {
        @include hires {
            @include sprite-map-background-size( $icons, .5 );
        }
    }
    .carousel-prev {
        background: sprite( $icons, arrow-left ) no-repeat;
        @include hires {
            background: sprite( $icons, arrow-left-2x, 0, 0, .5 ) no-repeat;
        }
    }
    .carousel-next {
        background: sprite( $icons, arrow-right ) no-repeat;
        @include hires {
            background: sprite( $icons, arrow-right-2x, 0, 0, .5 ) no-repeat;
        }
    }
}
```

This loads the hi-res sprite at a ratio of .5 (or 50%. both are valid) within my hi-res block:

``` scss
@mixin hires() {
    @media
        only screen and (min--moz-device-pixel-ratio: 1.5),
        only screen and (-ms-min-device-pixel-ratio: 1.5),
        only screen and (-o-min-device-pixel-ratio: 3/2),
        only screen and (-webkit-min-device-pixel-ratio: 1.5),
        only screen and (min-device-pixel-ratio: 1.5),
        only screen and (resolution: 300dpi)
    {
        @content;
    }
}
```

I will continue testing, but if you have a moment to take a look and let me know if I’ve missed anything crucial, I’d like to get this ready for prime-time and integration into Compass core.

PR #977, opened 2012-07-27
First blush attempt at addressing hi-res sprites

aarongustafson:
This is a re-post of Pull Request #970 from a new branch in my repo dedicated to this feature. I will close the other pull request momentarily.

I have added the .rb tweaks to my local Compass Gem (0.13 Alpha 0) and they seem to be working well. Here’s how the changes play out:

``` scss
.slidecontrols {
    a {
        @include hires {
            @include sprite-map-background-size( $icons, .5 );
        }
    }
    .carousel-prev {
        background: sprite( $icons, arrow-left ) no-repeat;
        @include hires {
            background: sprite( $icons, arrow-left-2x, 0, 0, .5 ) no-repeat;
        }
    }
    .carousel-next {
        background: sprite( $icons, arrow-right ) no-repeat;
        @include hires {
            background: sprite( $icons, arrow-right-2x, 0, 0, .5 ) no-repeat;
        }
    }
}
```

This loads the hi-res sprite at a ratio of .5 (or 50%. both are valid) within my hi-res block:

``` scss
@mixin hires() {
    @media
        only screen and (min--moz-device-pixel-ratio: 1.5),
        only screen and (-ms-min-device-pixel-ratio: 1.5),
        only screen and (-o-min-device-pixel-ratio: 3/2),
        only screen and (-webkit-min-device-pixel-ratio: 1.5),
        only screen and (min-device-pixel-ratio: 1.5),
        only screen and (resolution: 300dpi)
    {
        @content;
    }
}
```

I will continue testing, but if you have a moment to take a look and let me know if I’ve missed anything crucial, I’d like to get this ready for prime-time and integration into Compass core.

### DIFF
--- a/lib/compass/sass_extensions/functions/sprites.rb
+++ b/lib/compass/sass_extensions/functions/sprites.rb
@@ -1,5 +1,6 @@
 module Compass::SassExtensions::Functions::Sprites
   ZERO = Sass::Script::Number::new(0)
+  ONE = Sass::Script::Number::new(1)
   VALID_SELECTORS = %w(hover active target)
   # Provides a consistent interface for getting a variable in ruby
   # from a keyword argument hash that accounts for underscores/dash equivalence
@@ -57,7 +58,7 @@ module Compass::SassExtensions::Functions::Sprites
   # Becomes:
   #
   #     background: url('/images/icons.png?12345678') 0 -24px no-repeat;
-  def sprite(map, sprite, offset_x = ZERO, offset_y = ZERO)
+  def sprite(map, sprite, offset_x = ZERO, offset_y = ZERO, ratio = ONE)
     sprite = convert_sprite_name(sprite)    
     verify_map(map)
     unless sprite.is_a?(Sass::Script::String)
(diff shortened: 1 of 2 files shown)


lib/compass/sass_extensions/functions/sprites.rb
```diff
@@ -117,10 +119,14 @@ def sprite_has_selector(map, sprite, selector)
 
 
   # Returns a url to the sprite image.
-  def sprite_url(map)
+  def sprite_url(map, only_path = Sass::Script::Bool.new(false))
```
chriseppstein, line 122: I don't understand why this `only_path` argument is necessary.

lib/compass/sass_extensions/functions/sprites.rb
```diff
@@ -155,13 +162,19 @@ def sprite_position(map, sprite = nil, offset_x = ZERO, offset_y = ZERO)
...
+    else
+      r = ratio.value
+    end
     if offset_x.unit_str == "%"
-      x = offset_x # CE: Shouldn't this be a percentage of the total width?
+      x = (offset_x.value * r) # CE: Shouldn't this be a percentage of the total width?
+      x = Sass::Script::Number.new(x, x == 0 ? [] : ["%"])
     else
-      x = offset_x.value - image.left
+      x = ( offset_x.value - image.left ) * r
```
chriseppstein, line 174: I feel like it might be better to do this ratio manipulation in the sass mixins calling this function.

chriseppstein:
I have a very strong aversion to compass making delineations about what are the best resolution and media size breakpoints as I think these are a constantly moving target. At the very least, I would want to see this be configurable.

In particular, I really like [this post](http://mir.aculo.us/2012/06/14/more-than-meets-the-eye/) by @madrobby where he cautions against chasing the screen resolution. I think this post is sage advice and I want us to follow it.

It also seems just terrible to me that we would ask non-retina users to download a single sprite that is ~5x the non-retina size. It seems much better to have different sprite maps and just have the mixins select the right one.

If you'd like, please open a new issue and kick off a design descussion about what the right form this api and functionality should take.

borismus:
Agree that this should be configurable. Anything in the works here?

PR #978, opened 2012-07-27
Added support for display: (-prefix-?)inline-box;

aarongustafson:
The whole Flexbox implementation needs rewriting because [the W3C has changed the syntax](http://www.w3.org/TR/css3-flexbox/) (again). As I have time, I hope to work on it, but in the meantime, here’s support for box or inline-box.

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_box.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_box.scss
@@ -1,8 +1,13 @@
 @import "shared";
 
-// display:box; must be used for any of the other flexbox mixins to work properly
-@mixin display-box {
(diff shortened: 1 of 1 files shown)


chriseppstein:
Would love your help on this.

Regarding the display property. I think we should follow the pattern I've established in the images module where the mixin name matches the property name and compass looks at the values and prefixes them as necessary.

So ideally, this would become:

``` scss
@mixin display($value) {
  @if $value == box or $value == inline-box {
    @include experimental-value(display, $value,
      -moz, -webkit, not -o, -ms, not -khtml, official
    );
  } @else {
    display: $value;
  }
}
```

would you like to try this approach with flexbox module v2?

PR #986, opened 2012-08-09
Replace #root_footer with :after pseudo-element

krulik:
We can remove the extra markup element and have a cleaner HTML structure.
The downside is that a user-defined pseudo-element of the #root can interfere with the sticky-footer one.

### DIFF
--- a/frameworks/compass/stylesheets/compass/layout/_sticky-footer.scss
+++ b/frameworks/compass/stylesheets/compass/layout/_sticky-footer.scss
@@ -6,7 +6,7 @@
 //
 // Footer must be a fixed height.
 
-@mixin sticky-footer($footer-height, $root-selector: unquote("#root"), $root-footer-selector: unquote("#root_footer"), $footer-selector: unquote("#footer")) {
+@mixin sticky-footer($footer-height, $root-selector: unquote("#root"), $footer-selector: unquote("#footer")) {
   html, body {
     height: 100%; }
   #{$root-selector} {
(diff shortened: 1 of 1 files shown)


MoOx:
Maybe better to set #root_footer as :after by default, allowing people to set another element (to support IE7 ?)

krulik:
This is a good idea. However we will still need to add the `content: ""; display: block;` stuff for this to work.

Anahkiasen:
Well the "support for IE7" variable could be used no ?

MoOx:
I think so.

krulik:
I don't think you need anything explicit for IE7 - now it's `:after` by default but you can still overwrite it if you want to. It also doesn't break the mixin signature this way, much less obtrusive.

mirisuzanne:
I like this, but I think it belongs in the master branch, not the stable branch. It should also be documented, with tests, and added to the changelog. This is likely to affect everyone using the mixin, so we can't do it without warning.

krulik:
@ericam Fair, I'm new here and I feel that I need some contribution guidelines. Where can I find some? I've looked in the Wiki and the website and didn't find anything like that. How do I document/test/changelog? What branches are for what purpose? Sorry for n00b questions :)

mirisuzanne:
@krulik No worries. The contribution guidelines are on the site: http://compass-style.org/help/tutorials/contributing/

Let me know if you have more questions. Thanks for the contribution!

chriseppstein:
Closing due to lack of updates.

PR #996, opened 2012-08-20
change order of attributes for single-box-shadow to keep css standards

jcoufal:
I changed the order of single-box-shadow attributes in order to keep css standards (color afte #spread)

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_box-shadow.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_box-shadow.scss
@@ -52,11 +52,11 @@ $default-box-shadow-inset : false !default;
 // Provides a single cross-browser CSS box shadow for Webkit, Gecko, and CSS3.
 // Includes default arguments for color, horizontal offset, vertical offset, blur length, spread length, and inset.
 @mixin single-box-shadow(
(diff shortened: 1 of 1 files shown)


mirisuzanne:
This is highly backwards-incompatible. If we were going to make this change, it would have to happen in the master branch and get noted in the changelog.

Also, as I understand, the current order does _work_ in browsers, no? I'm -1 on it, but open to being convinced. Anyone else?

jcoufal:
I understand that it might be problem for backward compatibility. On the other hand it is important to keep standards and not "just use something what works".

Yes, it works in browsers.
- But one thing about current order is, that you don't have unified order for shadows across your mixins (single-text-shadow is different from single-border-shadow and I don't see why).
- Other thing is, as long as you keep standards and you want to move to Compass from some other tool (or redo your pure CSS with Compass) or vice-versa, you have to re-implement all box shadows.

What if we support current order as well as standard one? Because I would really vote for keeping standards as long as it makes sense.

webgefrickel:
I'm +1 on this. This was always a thing that made me use the non-single box-shadow mixin.

mirisuzanne:
I buy it. Please re-submit on the master branch, and add a note to the changelog.

PR #1007, opened 2012-08-31
Add selection mixin

corroded:
As @ericam suggested in the groups, here is the selection mixin using the new @content blocks from SASS 3.2

Hopefully this is what he meant :P 

_EDIT_
woops can someone delete/close this? I meant to PR on the master branch, not the stable one


scottdavis:
im going to close this because it needs to be rebased and re subbmited

mirisuzanne:
I just pulled in another version of this, so it's taken care of.

PR #1016, opened 2012-09-08
Added @experimental support for vendor-prefixed property and value.

dannygarcia:
There are some cases where prefixing both the property and value is useful.

One example is for creating a mask-image mixin that also prefixes the gradient:

``` scss
@mixin mask-image($gradient) {
   @include experimental-both(mask-image, $gradient); }
```


dannygarcia:
Another use case for this recently became apparent for implementing future-proof [transitions for CSS custom filters](http://alteredqualia.com/css-shaders/article/):

``` scss
-webkit-transition: -webkit-filter ease-in-out 1s;
```

Could be written as:

``` scss
@include experimental-both(transition, filter ease-in-out 1s);
```

scottdavis:
Needs a test case

chriseppstein:
The experimental mixin is basically unused now. Closing.

PR #1021, opened 2012-09-13
Feature/add documentation for selection mixin

corroded:
I also changed the implementation to use the parent selector for proper inheritance. Caveat is that you can't use the mixin alone(it's also included in the docs)


scottdavis:
@ericam whats the deal with this one?

mirisuzanne:
It needs to be cleaned up for more flexibility - same way #1044 was.

PR #1032, opened 2012-09-25
Add ability to set clear direction in pie-clearfix.

jfexyz:
Add argument to pie-clearfix mixin for clear direction. Defaults to 'both'.

Sometimes you need to be able to set this to not blow up floated layouts.


mirisuzanne:
Cool. I like this idea, and never even thought to try it. 

As a new feature I think it belongs in the master branch. Can you re-submit there? Thanks!

scottdavis:
Needs a test case

robwierzbowski:
I don't see a reason for adding this. Clearing the `:after` pseudoelement just forces the parent element to contain floats, and to do that effectively it should be set to `clear: both`. It should have no impact on the position of floated elements inside the parent element, and no impact on the position of the parent element itself if it is floated.

chriseppstein:
Agree with @robwierzbowski shouldn't be necessary. Please provide a clear use case if you want us to reopen.

PR #1033, opened 2012-09-26
Update frameworks/compass/stylesheets/compass/typography/lists/_inline-b...

emrecamasuvi:
...lock-list.scss

there is a strange bug here http://stackoverflow.com/questions/9664321/strange-margin-on-displayinline-block-elements
inline-block-list mixin needs to be updated.
font-size: 0 to ul to remove odd space and font-size: $value to display list item correctly.

### DIFF
--- a/frameworks/compass/stylesheets/compass/typography/lists/_inline-block-list.scss
+++ b/frameworks/compass/stylesheets/compass/typography/lists/_inline-block-list.scss
@@ -35,6 +35,12 @@
   @include no-bullet;
   @include inline-block;
   white-space: nowrap;
+  // needed to display list-items w/o odd space and correctly
+  @if $baseFontSize {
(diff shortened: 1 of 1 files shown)


mirisuzanne:
That's not a bug, that's how inline-block is meant to work according to the spec. I would consider an _optional_ white-space-removal technique, but I'm not interested in forcing that on users when they don't need white-space removed. This should be treated as a new feature in the master branch, not a bug fix in stable.

We could accept an optional white-space-font-size-fix argument, defaulted to `false`, and only trigger white-space removal if it is set. That also removes the need to guess a 12px font size.

madhaze:
has the optional white-space-font-size-fix argument been added? I could really use this.  Or is setting the font-size: 0; the only way to remove the white space between items?

scottdavis:
not that i know of

PR #1035, opened 2012-09-27
Added CSS3 input-placeholder() and input-placeholder-for-parent() mixins

borodean:


Anahkiasen:
Doesn't that one [already exists](https://github.com/chriseppstein/compass/blob/master/frameworks/compass/stylesheets/compass/css3/_user-interface.scss) ?

borodean:
Oh, you're right. However, there are no orphaned selector support implemented.

ry5n:
I'd rather avoid two separate mixins. I originally imagined that usage at the stylesheet root would look like this:

``` scss
* {
  @include input-placeholder {
    // rules
  }
}
```

I believe that is equivalent to using the bare pseudo-element. But it's not obvious that's how to do it. There are a couple of options if we want to keep a single mixin:

``` scss
@mixin input-placeholder($selector: "") {
  @if $experimental-support-for-webkit {
    #{$selector}::-webkit-input-placeholder { @content; }
  }
  // …
}
```

@param $selector could default to "" (empty) or "&". They optimize for different use cases and have different drawbacks. Using & is equivalent to the current mixin and doesn't work at root. Empty applies the rules to all _children_ of the current element, so doesn't work if you use it on the input itself. Anyone know how most people use this in the wild? Scoped to inputs, or bare?

mirisuzanne:
I think it's important that we support both, and I like the selector-argument approach. We have the same problem to solve with #1021, and they should be solved in the same manner.

ry5n:
Is there any way to tell from within a mixin whether it's being included at the stylesheet root or not? Could this be done in Ruby?

mirisuzanne:
Closed by #1044

PR #1043, opened 2012-10-07
Fix damaged images when creating projects on windows

jok2:
When a project is created on windows and an image is copied to the new 
directory, then the image is damaged because it is not read in binary
mode.


jok2:
Sorry for the additional commits, this is my first try with GitHub (and GitHub for windows)

Only the first commit is relevant. I tried compass on windows, and some images were damaged when i tried to create a project (even with the default frameworks). It happened with some .png images, the copied ones where one byte smaller than the original ones and couldn't be displayed with either some browser or an image viewer like irfanview.

PR #1044, opened 2012-10-08
Improve flexibility of input-placeholder mixin so it can be used at the ...

ry5n:
...root of the stylesheet or with an argument for use with a selector.

This creates two usage mode: at the stylesheet root, include the mixin with styles passed as mixin content. If a specific element is being styled, pass "&" as an argument to the mixin as well.

Addresses #1035, should be consistent with the solution for #1021. Note this changes the default behaviour of this mixin. Authors would need to update stylesheets, but this is not in stable yet, so that shouldn't be an issue.

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_user-interface.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_user-interface.scss
@@ -31,18 +31,18 @@
 //
 //     .textinput {
 //       @include input-placeholder("&") {
-//       	 color: #bfbfbf;
-//       	 font-style: italic;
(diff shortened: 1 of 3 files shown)


scottdavis:
ping @ericam this ok to merge?

ry5n:
Gave this an extra once-over and fixed some whitespace issues. Also added #1048 for consistency with this.

mirisuzanne:
I had no idea you could pass "&" as an argument and have Sass parse it. That's a very nice solution for this problem. I love it.

Issue #620, opened 2011-11-16
Sprites: New Function "sprite-names"


rvock:
Hi,

I would like to suggest a new function for the sprites:
sprite-names should return a (sass) list with all the (base)names of the files in a sprite for usage in other sprite-\* calls.

Then it would be possible to loop over them to use them in custom code or use the result to give to the sprites() function:

```
$sprite: sprite-map("icons/*.png");
@include sprites($sprite, sprite-names($sprite));

// OR:
@each $image in sprite-names($sprite) {
    .custom_prefix_#{$image} {
        background-position: sprite-position($sprite, '#{$image}');
    }
}
```

The Ruby Code to do this could look something like this: (I've never written Ruby before)

```
def sprite_names(map)
  Sass::Script::List.new(map.sprite_names.map { |f| Sass::Script::String.new(f) }, ' ')
end
```

Currently I have this code in a sass_additions functions file for my project :)

PR #1066, opened 2012-10-30
Adds reference for `sprite-names` sass function

chooh:
Refs #620


scottdavis:
I ended up adding this today not realizing this was pending open =(

PR #1081, opened 2012-11-02
Fix Thor tasks, add Rake tasks & update documentation

silvenon:
Thor commands had a **lot** of repetition in them and were kinda hard to use. Also, the documentation didn't mention the options, which were required, so instructed usage led to an error.

I removed the `--framework` option, because Blueprint is gone. Also, both `generate:example` and `generate:reference` take the same type of argument – `path/to/module`. For example, module `inline-block-list` would have the path `typography/lists/inline-block-list`.

This change is more dramatic when generating reference files, because before we had to do something like:

``` sh
$ ./bin/thor generate:reference compass/typography/lists/inline-block-list --stylesheet="../frameworks/compass/stylesheets/compass/typography/lists/_inline-block-list.scss"
```

Now we can do

``` sh
$ thor generate:reference typography/lists/inline-block-list
```

The rest is inferred. Templates generated by Thor are now much more realistic and complete, so minimal customization is required in the metadata.

Also, in both commands, Thor searches for the corresponding stylesheets and raises an error if it doesn't find them.

Thor tasks used to create `.sass` stylesheets for examples, now they create `.scss`. I converted the existing example stylesheets to `.scss` too.

Also, there's no need for `bundle exec` here, because `generator.thor` does bundler stuff inside it.

Lastly, I documented all of this stuff.

I rebased this onto the current master, so you should have no trouble merging.
# Wrong branch, sorry!


silvenon:
Holy sweet Jesus, wrong branch, sorry.

Issue #1002, opened 2012-08-24
Sprite image ordering


superbland:
Compass used create a sprite image, ordering the images by file name (I just upgraded, but sorry I'm not sure what version I was on previously). Although it was slightly hackish, if I wanted to have them be inserted in a particular order (e.g. to manually write background-position: 0px bottom;) I could simply add "z-" to the front of the file name and be sure it would be at the end of the file.

As far as I can tell, Compass now inserts files into a sprite based on file size, going from biggest to smallest. To me this makes less sense, and I would appreciate once again having some kind of control over the ordering, if I need it.

hagenburger:
I already implemented such a thing while I was experimenting with sprites for an iPhone website in normal and Retina resolution. Therefore it must have the same order. The code is here:

https://github.com/hagenburger/compass/commit/c472664b55269547de7231b7788ffb11c065e3ca

superbland:
Cool thanks, I'll check this out!

scottdavis:
@hagenburger did you want to make a PR for that change?

hagenburger:
It’s in https://github.com/hagenburger/compass/commit/c472664b55269547de7231b7788ffb11c065e3ca. It needs to be updated and tested. I can’t work on this right now as I’m both ill and in China :(

krisbulman:
This would be a nice improvement, doing the check for ordering differences between two sprite maps is not obvious if you need them to match (retina sprites) http://github.com/krisbulman/retina-sprite-mapper/blob/master/sass/_retina-sprites.scss#L94

robwierzbowski:
Related requests: https://github.com/chriseppstein/compass/issues/690

+1 from me. With predictable sorting based on names you can reduce the css needed for high pixel density image sprites to a background-image  and background-size declaration.

notethis:
+1

chrisforrette:
I'd like to +1 this as well.

I'm trying to generate a sprite for an animation sequence for a bunch of numbered files and it mashes them all together backwards.

scottdavis:
This is already implemented in master

skug:
As the work on v0.13 is still ongoing: is there a way to patch a vertical layout order by filename in the current stable release with an extension mechanism so it can be easily shared on different clients via project version control? You mentioned that it would be possible to write an own layout class in https://github.com/chriseppstein/compass/issues/690#issuecomment-8548102 but it seems, thats only the case with v0.13 as the base... 
[...]

PR #1083, opened 2012-11-03
sort sprites alphabetically

krisbulman:
This is basically @hagenburger's https://github.com/hagenburger/compass/commit/c472664b55269547de7231b7788ffb11c065e3ca on this issue #1002 as a pull request, with the optional fallback to width removed, as I don't think it's necessary.

### DIFF
--- a/lib/compass/sass_extensions/sprites/layout/vertical.rb
+++ b/lib/compass/sass_extensions/sprites/layout/vertical.rb
@@ -5,7 +5,7 @@ module Compass
         class Vertical < SpriteLayout
 
           def layout!
-            sort_by_width!
+            sort_by_name!
             calculate_width!
             calculate_positions!
             calculate_height!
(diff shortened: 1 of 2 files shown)


scottdavis:
We need to come up with another approach for this changing the sorting to alpha doesn't make sense for the people that specifically asked for width sorting. Sorting instead needs to be changed into something that is configurable.

scottdavis:
What nico had was acceptable as a patch I'm not sure why he never sent in a PR. There are reasons for sorting by width and file size so that you smaller images become available in progressive loading in browsers that support it.

hagenburger:
@scottdavis just came back from my two months China trip few hours ago. What I did worked quite well, but without any test. I would have a look at it again at the weekend.

scottdavis:
No biggie im cooking up something now that may be more configurable

scottdavis:
besides alpha sorting and size sorting what else should be offered.

notethis:
What's the status of sorting by alpha?

sbdev:
This sort-by parameter would be a great way to optimize css dowloads.
CSS files that don't use newly added icons currently are likely to require new offsets and therefore require version bump and re-download by all users :-(

https://github.com/chriseppstein/compass/commit/520c950f7d9a9bf7d5cf99842e898feec8aa7364
Why it didn't make it into 0.12.2?

scottdavis:
Mostly because 0.12.2 was released way before that change was made

sbdev:
oops didn't realize 0.12.2 is one-year old (06/24/2012). Sorry!
Any plan for a 0.12.3?

PR #1095, opened 2012-11-14
added Sprite Size functions

cwerner1:
To Recieve the Dimensions of a Sprite I added the functions 
sprite-height,sprite-width and sprite-size

I use it to Resize the "Retina" sprite to its original size


scottdavis:
This has already been added in the master branch.
https://github.com/chriseppstein/compass/blob/master/lib/compass/sass_extensions/functions/sprites.rb#L14-27

PR #1100, opened 2012-11-21
Fix images_path bug in the image cache buster (Fixes #1092)

### LINKED ISSUE #1092: images_path not being honored
In Compass 0.12.2 setting the `images_path` in config.rb has no effect. I receive warnings that image foo can not be found in <current directory>/images. If I change images_dir, it will look to that relative directory, but never to the full path. What am I missing?

acasademont:
As said in #1092, images_path is not being honored and the image cache buster fails as it can't find the image

### DIFF
--- a/lib/compass/sass_extensions/functions/urls.rb
+++ b/lib/compass/sass_extensions/functions/urls.rb
@@ -103,8 +103,10 @@ module Compass::SassExtensions::Functions::Urls
       end
 
       # Compute the real path to the image on the file stystem if the images_dir is set.
-      real_path = if Compass.configuration.images_dir
-        File.join(Compass.configuration.project_path, Compass.configuration.images_dir, path)
+      real_path = if Compass.configuration.images_path
+        File.join(Compass.configuration.images_path, path)
+      else
+        File.join(Compass.configuration.project_path, path)
       end
 
       # prepend the path to the image if there's one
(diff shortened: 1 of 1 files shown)


acasademont:
@scottdavis do you need something more for this PR? Thanks!

danielguillan:
+1

lib/compass/sass_extensions/functions/urls.rb
```diff
@@ -103,8 +103,10 @@ def image_url(path, only_path = Sass::Script::Bool.new(false), cache_buster = Sa
       end
 
       # Compute the real path to the image on the file stystem if the images_dir is set.
-      real_path = if Compass.configuration.images_dir
-        File.join(Compass.configuration.project_path, Compass.configuration.images_dir, path)
+      real_path = if Compass.configuration.images_path
+        File.join(Compass.configuration.images_path, path)
+      else
+        File.join(Compass.configuration.project_path, path)
       end
```
scottdavis, line 110: I'm not sure this change is going ot give you what you are expecting. Why would you store your images in the root of the project path?

acasademont, line 110: The thing is that I prefer to use absolute paths and here the absolute "image_path" setting was not being honored. In fact, this is just the same snippet of code that you did in, which is perfect.

https://github.com/chriseppstein/compass/blob/master/lib/compass/sass_extensions/functions/image_size.rb

scottdavis, line 110: ok

scottdavis:
4912c73..3bab458

PR #1106, opened 2012-11-27
Add support for positioning sprites with percentages

trotzig:
The default behaviour for sprites is to position the background image
with pixels. This works if the sprite itself will always be displayed in
its original pixel dimensions. However, as more and more websites turn
responsive, is is vital that we can scale images up and down (mostly
down).

This commit will add support for positioning the background image with
percentages instead of pixels. This will make it easier to scale a
sprite image up and down. You won't be able to do it out of the box
though, you still need to do some manual work:
- Using "background-size: cover" instead of the default (auto)
- Placing the sprite in a container with a specific width, then using
  the following magic to the sprite:
    height: 0;
    padding-top: 75%; //image height/width ratio
    width: 100%;


ghost:
This would be useful for us.

scottdavis:
I hate to ask this since it looks really good but, can you rebase against master I'm afraid this will not merge cleanly from stable -> master

trotzig:
No problems. I'll get back to you as soon as I can.

trotzig:
I've rebased this onto the master branch. I didn't find a way to edit this pull request to point to a different base so I created a new pull request: https://github.com/chriseppstein/compass/pull/1112

sam3k:
how do i use this?

PR #1120, opened 2012-12-17
Added 'text-transform:capitalize' to 'hide-text' mixin

clark-pan:
Added 'text-transform:capitalize' to the 'hide-text' mixin to fix issue with this text-replacement technique in IE7 on any <input> elements.

Here's a test case: http://jsbin.com/olikuj/1

Tested in ie7 with browserstack.


chriseppstein:
Merged to master. This property will only be output when IE7 is a supported browser (See: db06ca838b7a6765424df87336f275d73c5e8e20)

PR #1129, opened 2013-01-03
Flush the stdout buffer after compiler runs

leik:
Flush the stdout buffer after compiler runs to display all the standard outputs correctly when running in NodeJS spawn. Otherwise it outputs weird result:

When running "compass watch" by Node.js child_process.spawn like this :

``` javascript
var compass = require('child_process').spawn('compass', [ 'watch']);
compass.stdout.on('data', function (data) {
    console.log('' + data);
});
```

For the first time it only outputs : ">>> Change detected at 16:34:20 to: XXX", the expected followed lines are missing until another change is made.


scottdavis:
I "think" this is fixed in master

scottdavis:
yes this entire file has been rewritten on master for the next release. So this is invalid but, thank you =).

PR #1133, opened 2013-01-07
Use image-url instead of stylesheet-url for ellipsis.xml

atambo:
This is because stylesheet-url automatically appends .css to the
end which makes the url look like:

xml/ellipsis.xml#ellipsis.css

when it should look like:

xml/ellipsis.xml#ellipsis


atambo:
This probably isn't the right way to fix this. Is there anyway for me to tell stylesheet-url to not append .css to the end of the asset? I'm not even sure where that .css is coming from.

scottdavis:
I'm not seeing this problem? ive written a test for it and it passes. 714489e3d86aab953fde2999054d0cf274cf80d6

PR #1137, opened 2013-01-11
Enable cache busting on font-url helper.

marshallbrekka:
I've set the default value for `$cache-buster` to false so to keep it backwards compatible.


test/fixtures/stylesheets/busted_font_urls/config.rb
```diff
@@ -0,0 +1,29 @@
...
+# http_images_path = :relative
+http_fonts_path = "/fonts"
+line_comments = false
+
+asset_cache_buster do |path, file|
+  pathname = Pathname.new(path)
+
+  case pathname.basename(pathname.extname).to_s
+  when "grid"
+    new_path = "%s/%s-BUSTED%s" % [pathname.dirname, pathname.basename(pathname.extname), pathname.extname]
```
scottdavis: I really hate this type of string formatting and in the past we have rejected PR's because of it please change to "#{}" method

marshallbrekka: will do. Just FYI this is taken directly from the busted_image_urls/config.rb file

scottdavis:
I cherry-picked this into master. everything was good except your testcase was incorrect and failing see: ba1bbfa26044accd4eea238d5363c251bb10f01a

PR #1140, opened 2013-01-16
Uses :filesystem_importer sass option, when provided, to create importers

svallory:
Sass allows the user to overwrite the default importer using :filesystem_importer option. This makes Compass respect that option when used.

I didn't added any tests because, honestly, I don't think I could do a decent job without putting a LOT of effort to it, and this is a very small modification.

I'm starting a new OSS project called Liftr and I created a custom importer for sass which only imports a file once (https://gist.github.com/4549615) but I need to be able to set the `:filesystem_importer` option (supported by sass) for it to work. I think this import would actually be a nice addition to compass or even sass.

### DIFF
--- a/lib/compass/compiler.rb
+++ b/lib/compass/compiler.rb
@@ -15,7 +15,8 @@ module Compass
       self.sass_options.delete(:quiet)
       self.sass_options.update(sass_opts)
       self.sass_options[:cache_location] ||= determine_cache_location
-      self.sass_options[:importer] = self.importer = Sass::Importers::Filesystem.new(from)
+      self.sass_options[:filesystem_importer] ||= Sass::Importers::Filesystem
+      self.sass_options[:importer] = self.importer = self.sass_options[:filesystem_importer].new(from)
       self.sass_options[:compass] ||= {}
       self.sass_options[:compass][:logger] = self.logger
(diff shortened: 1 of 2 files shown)


scottdavis:
I am hesitant to merge this without a test case small or not

svallory:
@Scottdavis, I understand, I do believe they are important and, in this case, fairly easy to write. But I wouldn't know how to do that properly unless a learned a lot more about compass.

Do you think you could write those tests? Or ask someone else to do it? Or give me a little guidance on how to properly check if compass is doing all it needs to do with this new setting?

svallory:
Hey @scottdavis, there is already a test `test_sass_options` in [test/units/configuration_test.rb#L297](https://github.com/chriseppstein/compass/blob/stable/test/units/configuration_test.rb#L297) which ensures I can set a sass option through configuration and access it later. Do you think we need to test anything else?

svallory:
@scottdavis any thoughts on my last comment?

svallory:
Can someone please consider this? @scottdavis ? It already has a test covering this change at [test/units/configuration_test.rb#L297](https://github.com/chriseppstein/compass/blob/stable/test/units/configuration_test.rb#L297)

Build is failing because of unrelated tests. I can update it to the current version of compass, but I'm hesitant to do this and simply waste my time.

PR #1148, opened 2013-01-25
added $spacing variable to sprite-map

micahgodbolt:
The $spacing variable was not being passed to the sprite-map function, so setting $name-spacing had no effect.


scottdavis:
Already fixed on master

PR #1150, opened 2013-01-26
- 1 param, + pseudo elem instead of div

nissoh:


scottdavis:
Can you be a bit more descriptive on that this change is? also this will need the test case updated.

PR #1155, opened 2013-01-30
Make the project_path variable setable as an option on CLI.

androa:
This makes Compass a lot easier to use in automated deployment processes.


scottdavis:
why `app-dir` and not `project-path`?

androa:
Yeah, I thougth to use project-path as well, but that would be inconsistent of the similar --app = project_type option. I see now that I've messed up something in my branch after sending the PR, I can try to clean that up when I'm done snowboarding.

androa:
So, there I cleaned up the git history.

PR #1161, opened 2013-02-05
Update lib/compass/sass_extensions/functions/gradient_support.rb

dalgard:
Correction: Resulted in a gradient 100px wide instead of 100%

### DIFF
--- a/lib/compass/sass_extensions/functions/gradient_support.rb
+++ b/lib/compass/sass_extensions/functions/gradient_support.rb
@@ -351,7 +351,7 @@ module Compass::SassExtensions::Functions::GradientSupport
     # returns the end position of the gradient from the color stop
     def grad_end_position(color_list, radial = Sass::Script::Bool.new(false))
       assert_type color_list, :List
-      default = Sass::Script::Number.new(100)
+      default = Sass::Script::Number.new(100, ["%"])
       grad_position(color_list, Sass::Script::Number.new(color_list.value.size), default, radial)
     end
 


scottdavis:
You doing to have to be more detailed about what caused this change and what the expected output was. The fact that there is not a test case makes me nervous.

dalgard:
Sorry for the delay. The module generates the SVG based gradients used for (experimental) compatibility with IE9 and less. Radial-gradients is a special case; it is probably not used much, which must be why this very small error has gone unnoticed.

As you can see in the diff, I've just made a tiny correction - it is _totally clear_ from the code that this was supposed to be a percentage, like in the other routines, and the change is _only_ reflected in the SVG string directly (which has that character added, making the gradient fill up the entire image instead of an absolute value of 100px, which doesn't make sense).

If you ask the person who has written this part of the code, they will tell you right away that this is how it's supposed to be. This may be the most trivial correction ever submitted on GitHub and as a contribution it looks kinda silly, but it did help me in my project :)

andymesa:
I'm running into this problem and can confirm that dalgard's patch fixes it.

scottdavis:
This has been rewritten in the latest alpha

andymesa:
Do you mean in the next alpha? Because it's still an issue in 0.13.alpha.4 (that is fixed by adding back the %)

Also, thanks for the quick response!

scottdavis:
Ok

chriseppstein:
I think this patch should have fixed it: https://github.com/chriseppstein/compass/pull/1436

it landed a couple days ago. Please confirm. closing for now.

PR #1163, opened 2013-02-07
Update Sass dependency to 3.2.5

alecperkins:
Sass 3.2.0.alpha.93 doesn't support variable arguments, so relying on Compass to install Sass results in syntax errors inside Compass, eg when importing [css3](https://github.com/chriseppstein/compass/blob/1724d70bd9c77aff31dc1c6e596e8b476fc7a656/frameworks/compass/stylesheets/compass/css3/_shared.scss#L220).


Anahkiasen:
+1, ran into this a few times.

scottdavis:
Fixed on master cd6028d

PR #1166, opened 2013-02-11
Do not overwrite default value for mozilla legacy support.

dvessel:
Sorry for the [similar pull](1165). I wasn't sure which branch to apply this to.

I need to disable this flag.


scottdavis:
merged in #1165 to master

PR #1169, opened 2013-02-13
background-size doesn't need prefixes for opera or webkit

lkrids:
background-size doesn't need prefixes for opera or webkit: http://www.standardista.com/css3/css3-background-properties

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_background-size.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_background-size.scss
@@ -22,5 +22,5 @@ $default-background-size: 100% auto !default;
 ) {
   $size-1: if(type-of($size-1) == string, unquote($size-1), $size-1);
   $sizes: compact($size-1, $size-2, $size-3, $size-4, $size-5, $size-6, $size-7, $size-8, $size-9, $size-10);
-  @include experimental(background-size, $sizes, -moz, -webkit, -o, not -ms, not -khtml);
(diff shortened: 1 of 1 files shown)


lkrids:
http://caniuse.com/#search=background-size

scottdavis:
The tests need to be updated also

Anahkiasen:
I don't think Firefox needs the prefix either, as shown on the link above, no ?

chriseppstein:
``` scss
@import "compass/css3";

$debug-browser-support: true;
.bs {
  @include background-size;
}
```

is compiled to:

``` css
/* line 4, ../sass/screen.scss */
.bs {
  /* Capability background-img-opts is prefixed with -moz because 0.27123% of users need it which is more than the threshold of 0.01%. */
  /* Creating new -moz context. */
  -moz-background-size: 100% auto;
  /* Capability background-img-opts is not prefixed with -ms because 0% of users are affected which is less than the threshold of 0.01. */
  /* Capability background-img-opts is prefixed with -o because 0.02466% of users need it which is more than the threshold of 0.01%. */
  /* Creating new -o context. */
  -o-background-size: 100% auto;
  /* Capability background-img-opts is prefixed with -webkit because 1.94978% of users need it which is more than the threshold of 0.01%. */
  /* Creating new -webkit context. */
  -webkit-background-size: 100% auto;
  background-size: 100% auto;
}
```

You can set your `$background-size-threshold` to `2` if you really want to omit it for the 2% of webkit users that still need it. If you set it to `0.05` it would prune out just the opera prefix. If you set it to `0.3` it would prune out opera and firefox.

PR #1177, opened 2013-02-19
Loosen `listen` gem dependency

jsmestad:
See the commit message for details. Guard makes use of some new methods in the `listen` gem. In order to use both 1.6.x of `guard` and `compass` we need to loosen the dependencies a bit on versions.

Ended up just doing a range (it may work with latest listen gem `0.7.x` but did not want to mess testing every version and instead just fix the specific issue.

`gem.add_dependency 'listen', ['>= 0.5.3', '<= 0.7.0']` is what the line now looks like.

### DIFF
--- a/compass.gemspec
+++ b/compass.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |gemspec|
 
   gemspec.add_dependency 'sass', '~> 3.2.5'
   gemspec.add_dependency 'chunky_png', '~> 1.2'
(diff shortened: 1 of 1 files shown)


jjb:
@scottdavis @ericam @jsmestad now guard calls for listen `>= 1.0.0` -- can compass work with this as well or has the listen api changed?

scottdavis:
https://github.com/guard/listen/issues/87

jjb:
@scottdavis so it looks like 1.8 support was added back in before 1.0... so is there still an obstacle to requiring `< 1.1.0` ?

PR #1180, opened 2013-02-21
README: update Code Climate badge

artagnon:
The new badge also has an overall GPA attached to it.


scottdavis:
merged into master

Issue #1184, opened 2013-02-25
Compiling a single file will always rewrite Sprite Images


michaelgilley:
When compiling a single file (`$ compass compile [/path/to/file.scss]`) the --force flag is automatically set to true. This causes all sprite files to be regenerated unnecessarily causing longer compile times.

See 6d6e1ca0e3d97ee7758ccdebc5c6ec0558d2fa7c.

michaelgilley:
I'd suggest this not be implied. If someone needs to overwrite the file then they should explicitly use the `--force` flag. E.g., `$ compass compile myfile.scss --force`.

chriseppstein:
This seems fine to me, can you make a PR against stable?

michaelgilley:
Done.

bdkjones:
Question: what are the implications of this change? Do I need to ensure that the CSS file does not exist before calling:

```
compass compile [path/to/file.scss]
```

Or will Compass still overwrite any previous CSS file that exists at the output location for this file?

michaelgilley:
Compass will write/overwrite the CSS file as expected but only IF the scss
fill has been changed. Previously, it was overwriting regardless of no
changes.

bdkjones:
How does Compass know if the file's been changed if I'm not using the --watch command? Does it keep a cached version of the file's last state?

michaelgilley:
Yes. I'm not familiar enough to comment on how this mechanism works
(haven't dove into it) but compass stores a cached version of each file in
the .sass-cache dir.

michaelgilley:
Yes. I'm not familiar enough to comment on how this mechanism works
(haven't dove into it) but compass stores a cached version of each file in
the .sass-cache dir.

Chris can speak more to this.

I have noticed, though, whenever I edit a file that will compile to the
same result (such as add or remove comment lines) the output in the CLI
looks like this:

$ compass compile
identical assets/css/main.css

PR #1186, opened 2013-02-26
Force should be explicitly stated in call.

michaelgilley:
Automatically setting the force flag to true will cause all sprite maps generated by a compiling file to also be regenerated, even if they haven't been changed, in turn causing compilation to be overly slow. This was especially the case in third party programs that watch project directories on their own and call compass compile on individual files (instead of assuming `compass watch`).

Instead use `compass compile my_file.scss --force`.


michaelgilley:
See related https://github.com/chriseppstein/compass/issues/1184

scottdavis:
can you re-base this change against the master branch please.

PR #1191, opened 2013-02-28
Sprites define background image and repeat separately

pivotal-medici:
Depending on structure SASS may split a sprite's `background` and `background-position` into different selectors. After processing the selector containing `background` may come after the selector with `background-position`. When this happens `background`'s default background-position value (`0% 0%`) takes precedence.

Currently background image and repeat are defined in a composite property. This patch splits it into `background-image` and `background-repeat` to avoid the aforementioned problem.


scottdavis:
I'm ok with this if @hagenburger is.

pivotal-medici:
Sweet! Thanks @scottdavis

Issue #921, opened 2012-06-05
support for new flexbox spec


chriseppstein:
There's a new spec for flexbox. We need to sort out what that means and how to support it going forward.

chriseppstein:
According to @tabatkins The official docs are still a work in progress and there's no conforming implementation. So we'll wait a bit to implement this.

https://twitter.com/chriseppstein/status/210119225031933954

tabatkins:
Specifically, give it two to three months.  Spec'll be CR by then, and we should have just started shipping unprefixed, alongside Opera and Firefox.

The only prefix you'll need to worry about in a few months is MS, because the spec stabilized just a little bit too late to make their IE10 cutoff.

(Unless you want to hack support via old Flexbox (the box-\* stuff), which is possible for some scenarios, but not all.)

alanhogan:
> work in progress and there’s no conforming implementation

Hmm, does that mean the [CanIUse flexbox page](http://caniuse.com/#feat=flexbox) is wrong/out of date for showing lots of Chrome versions in green?

Personally, I would definitely like to see a flexbox mixin that uses the older version when possible. If I have some simple layout issues that could benefit from this sort of thing (and I do), then in the spirit of Compass, I want to declare it once and have it work no matter which browser I’m in (assuming it’s reasonably possible in pure CSS in that browser, natch). I’m okay with a less functional mixin that meets those requirements as opposed to a fully-functional latest-spec-only variant (though both probably ought to exist, so authors can do what they will).

tabatkins:
Yes, the page is wrong.  It's mixing together both the 2009 version of flexbox (where all the properties had the box-\* prefix), and the current one.

At the moment, only Chrome has an implementation of new Flexbox, and we're about to ship breaking changes to make the naming match the Last Call draft that was published today.  Very shortly (in a few months), Chrome, Firefox, and Opera will all be shipping unprefixed Flexbox.  IE10 will also ship Flexbox, but possibly with a prefix.  Obviously, some mixins that paper over the temporary -ms- prefix (if it ends up existing) will be nice.

It is definitely possible to do a lot of the new draft's stuff with the old draft's syntax, so mixins that do so would be good.  The spec has changed very little in overall concept since 2009; the new spec just has a lot of naming changes and a _lot_ of cleanup for usability and detail.

alanhogan:
Fascinating. Thanks.

paulirish:
[...]

PR #1195, opened 2013-03-03
Stable

cimmanon:
Added new mixins for Flexbox as referenced from https://github.com/chriseppstein/compass/issues/921.  Left out the typical `$default-property-name` style variables because I worry about a potential naming conflict regarding the `$default-order` variable to go with the `order` mixin.

Added test for new mixin collection (I hope I've done this right)

Removed -ms- and standard prefixes from the old box collection.  IE10 doesn't use the draft these properties come from.

Updated test for old box collection

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_box.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_box.scss
@@ -1,6 +1,4 @@
-@import "compass/css3/shared";
-
-//@import "shared";
+@import "shared";
 
 // display:box; must be used for any of the other flexbox mixins to work properly
 @mixin display-box {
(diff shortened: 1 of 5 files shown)


mirisuzanne:
Thanks! Looks great! I think this should be pulled into the master branch. Can you re-request over there?

I also think the "$default-" variables would be a good idea, just with a bit of name-spacing. How do you feel about something like `$default-flexbox-order` or `$default-flex-order`?

Thanks again for the work. This is awesome.

cimmanon:
If one of them is namespaced, then they all need to be namespaced for consistency.  I couldn't find any other module where the default variable didn't match the property name, so I think namespacing would lead to confusion because it breaks expectations.  For what it's worth, the Columns module doesn't include default variables, even on properties where it might make sense (column-rule, column-gap).

If they must be added, then `$default-flex-order` sounds like the better choice to me.  I'll resubmit to master a little later today.

PR #1199, opened 2013-03-04
Fixes #941, ellipsis support for IE

### LINKED ISSUE #941: IE 8/9, ellipsis, and word-wrap
IE 8/9 won't respect the `white-space: nowrap` declaration in the ellipsis mixin when there is a parent that has `word-wrap: break-word`. The result is that no ellipsis is shown and the text will get broken and wrap across multiple lines.

The fix is simple, add `word-wrap: normal` to the ellipsis mixin to act as a guard against this IE 8/9 behaviour. This seems like it would be a fairly safe thing to do, but could benefit from a little more exploration.

Test case & fix: http://jsfiddle.net/P2v4c/

(Thanks to everyone below who helped identify the extent of this bug in various versions of IE)

makangus:
Added word-wrap for ellipsis support for IE

### DIFF
--- a/frameworks/compass/stylesheets/compass/typography/text/_ellipsis.scss
+++ b/frameworks/compass/stylesheets/compass/typography/text/_ellipsis.scss
@@ -11,6 +11,7 @@ $use-mozilla-ellipsis-binding: false !default;
 @mixin ellipsis($no-wrap: true) {
   @if $no-wrap { white-space: nowrap; }
   overflow: hidden;
+  word-wrap: normal;
   @include experimental(text-overflow, ellipsis,
     not -moz,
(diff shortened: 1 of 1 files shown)


mirisuzanne:
Thanks! Can you add tests for this, please?

chriseppstein:
https://developer.mozilla.org/en-US/docs/Web/CSS/word-wrap?redirectlocale=en-US&redirectslug=CSS%2Fword-wrap

There's several changes needed here. The property suggested is actually for legacy support, the new property is `overflow-wrap`. So we need to specifically target the legacy browsers.

chriseppstein:

PR #1201, opened 2013-03-06
Update transform module

mirisuzanne:
This all seems to be working, and the tests all pass. I would just like some other people to look over it before we pull it in. It includes a lot of big changes to the module.


will-hancock:
So a fresh install of compass on my new PC is returning: 

"Line 573 of _transform.scss: Mixin transform-origin takes 1 argument but 4 were passed."  

I've checked all the mixins, I'm using in the project, all our correct... It must be other mixins, using mixins using mixins.

How would i trace that?

chriseppstein:
this was re-written for 1.0 already.

PR #1202, opened 2013-03-06
Transitions WIP

mirisuzanne:
I updated the basic browser support matrix, and removed the crazy-old webkit workaround. I also removed the old code for prefixing transitioned properties, because it was over-simplistic. 

It would be nice to add that back in, with a more robust approach, but I couldn't figure out a way to make it happen. Help, @chriseppstein or @scottdavis ?

I did update the tests, so they fail currently, but should pass if you get it working. I also updated `$transitionable-prefixed-values` so it shows a list of values and the prefixes that should be applied to them.

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_transition.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_transition.scss
@@ -1,19 +1,11 @@
+// CSS Transitions
+
 @import "shared";
 
-// CSS Transitions
-// Currently only works in Webkit.
-//
-// * expected in CSS3, FireFox 3.6/7 and Opera Presto 2.3
-// * We'll be prepared.
-//
-// Including this submodule sets following defaults for the mixins:
-//
-//     $default-transition-property : all
-//     $default-transition-duration : 1s
-//     $default-transition-function : false
-//     $default-transition-delay    : false
-//
-// Override them if you like. Timing-function and delay are set to false for browser defaults (ease, 0s).
+// @private css3-feature-support variables must always include a list of five boolean values
+// representing in order: -moz, -webkit, -ms, -o, -khtml.
+$transition-support: not -moz, -webkit, not -ms, not -o, not -khtml;
+
 
 $default-transition-property: all !default;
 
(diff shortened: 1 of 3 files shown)


robwierzbowski:
What's the desired prefixing behavior? I will gladly help any patch with a 5/1 deletions/additions ratio.

mirisuzanne:
`$transitionable-prefixed-values` should contain all the details. It's a nested list of properties and the prefixes required for each property. For example, the first value (`transform ms webkit`) means when I do this:

``` scss
.transition {
  @include transition(transform 300ms);
}
```

It should output this:

``` scss
.transition {
  -webkit-transition: -webkit-transform 300ms;
  transition: -ms-transform 300ms;
  transition: transform: 300ms;
}
```

Does that make sense?

robwierzbowski:
Sure does. Code forthcoming.

robwierzbowski:
Spent some time in the code and understand the situation now. I can do this as a one off in transitions, but it might be better to add an `experimental-property-value()` mixin to the `experimental()` and `experimental-value()` mixins in `_shared.scss`. Something like:

``` scss
@mixin experimental-property-value($property, $value, [$prefixes]...) {
  ...
}
```

where each `$prefixes` is `$prefix-property $prefix-value`. Usage would be:

``` scss
.test {
  @include experimental-property-value(
    transition, 
    transform, 
    webkit webkit, default ms, default default);
}
```

If we add this, `experimental()` and `experimental-value()` could be simple wrappers to `experimental-property-value()`. Thoughts?

mirisuzanne:
I do like the idea of doing it generically. I'm not sure I understand the full extent of the proposal, but I'd love to see it. :)

robwierzbowski:
Alright, coming up!

robwierzbowski:
Well that got a little messy, but the code is up now.

chriseppstein:
I rewrote this file as part of 1.0. Would love it if you would review the code and let me know if there's any issues.

chriseppstein:
New code is here: https://github.com/chriseppstein/compass/blob/master/core/stylesheets/compass/css3/_transition.scss

robwierzbowski:
I haven't looked at 1.0 yet in depth so the patterns aren't really familiar to me, but looks much cleaner than what we had. Code comments are A+ #1. Ship it!

PR #1214, opened 2013-03-14
Do not set --force when passing files as arguments

fabiomcosta:
There is no need to set force to true here since compass already
has a --force true option and sass/compass does a very good job on
detecting files that changed.

My use case for using compass with django-pipeline, in which it will
force compilation every request while developing (DEBUG=True there).

Its not possible to use `compass watch` because the files are going to
be inside the installed apps, that could be anywhere, even outside
your project folder.


scottdavis:
This is already fixed in master. I thought I had Déjà vu

scottdavis:
f87c47d7d0886bd75cb0ad58194499fee28533dd

fabiomcosta:
Nice! thank you for the quick response.

PR #1220, opened 2013-03-19
Full height sticky footer

alxjrvs:
I've added functionality to the sticky-footer class that allows for full-height elements that will stretch the full length of the document, ending wherever the sticky footer begins. 

It leverages the calc() css function and relies on a relatively positioned root, the latter of which I have kept out of this so as to avoid magic behavior. Since you all have a much more top-down idea of how compass works, I'll leave it to your discretion as to whether or not it should be default. 

With this change, you can make any element on the page stretch from the Top of the root to the sticky footer by adding the class '.full_height'.

### DIFF
--- a/frameworks/compass/stylesheets/compass/layout/_sticky-footer.scss
+++ b/frameworks/compass/stylesheets/compass/layout/_sticky-footer.scss
@@ -6,7 +6,7 @@
 //
 // Footer must be a fixed height.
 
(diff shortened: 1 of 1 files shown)


chriseppstein:
It's a nice pattern, but I'm not going to add it because I'm trying to pivot compass to being less about "all the cool css tricks" and more about "all the cool sass tricks".

Issue #1221, opened 2013-03-19
input placeholder mixin not working in firefox 19+


rzschoch:
The current mixin only outputs :-moz-placeholder which isn't working in the more current versions of Firefox. Since Firefox 19+ this attribute should be referenced as an pseudo class (prefixed with ::) and not longer as an pseudo element (prefixed with :).

chriseppstein:
fixed in 1.0

PR #1222, opened 2013-03-19
Fix for issue #1221

rzschoch:
I added the placeholder pseudo class to support newer versions of firefox and modified the related test fixture.


chriseppstein:
The work done on the new prefixing system in 1.0 added this.

PR #1230, opened 2013-03-26
fix custom watch function

tka:
block as parameter need '&'


scottdavis:
nice catch i prob should add a regression test for this before merging

PR #1247, opened 2013-04-10
Add inverse trigonometry functions

AndrewO:
I was working with some css image gradient backgrounds the other day and found the need to do some trigonometry. To my delight, I saw that sin, cos, and tan were available in Compass. But then I realized I actually needed the inverse trig functions: asin, acos, and atan.

Looking at the source, they seemed easy enough to add, so rather than complain about it, I actually did it. Please accept my patch so that I can do all sorts of crazy things that probably won't work in IE for like 3 more versions. Thanks.


scottdavis:
neato merged!

PR #1249, opened 2013-04-11
Update Sprites _base.scss stylesheet

MPSinclair:
Add additional selector class based on selector name; typically used with state changing via Javascript


scottdavis:
This will break all the sprite test cases they will need to be updated

chriseppstein:
I'd rather not add this for everyone. There's so many selectors already. This can be done in your project by writing your own selectors or overriding our mixin.

Issue #1051, opened 2012-10-15
more robust color contrast mixin


bluesaunders:
Hello,

I recently [tweeted](https://twitter.com/Blue642/status/257510364307152897) about a color contrast mixin that I authored based on a W3C algorithm for color contrast. [(Article link...)](http://www.w3.org/WAI/ER/WD-AERT/#color-contrast) I was asked to make an issue here, discussing the technique.

Compass's [color contrast module](https://github.com/chriseppstein/compass/blob/stable/frameworks/compass/stylesheets/compass/utilities/color/_contrast.scss) seems to favor dark text because it depends on the HSLA lightness of a color. Selecting between light and dark text based on a lightness threshold is not enough to make proper color contrast decisions.

I've used the algorithm here for finding the difference between a given background color, and specified light, and dark values. The mixin will use the light or dark value that has a greater difference in contrast form the background color.

So, instead of relying on a threshold of lightness, the result here is that you always get the value that is higher contrasted with the background color.

NOTE: The goal was to provide the most contrasted color between the light and dark values supplied. So it is possible to have low contrasting colors, if the light or dark values are not much different than the background.

You can see my results [here](http://codepen.io/bluesaunders/pen/FCLaz).

PS. I left the compass "contrasted" mixin commented in the code, un-comment to see the difference between contrast choices.

ry5n:
I don't use the contrast module, but lately I keep thinking I should. This seems to be a clear improvement in the results. Well done!

I think the core team would welcome a pull request :)

bluesaunders:
Thanks! I'd like to see a little discussion around this before I make a pull request, but if it's well-received, I'd be more than happy to make my first contribution to Compass :)

---

Here is a slightly cleaner version of the example that more closely follows the structure of the existing color contrast module. http://codepen.io/bluesaunders/pen/bEkuh

Edit: i did some digging and discovered that originally the the contrast utility did something similar to what I'm doing (using the same "YIQ" algorithm) but was later switched to using a lightness threshold instead. [see here](https://groups.google.com/forum/?fromgroups=#!searchin/compass-users/contrast/compass-users/DHYbWbttANc/wpaDIGZbwAkJ)

The technique I'm using is a bit different than the original in that, instead of checking against a set range, I look at the absolute values of the differences in contrast and choose the higher number.

scottdavis:
I would welcome this on the grounds that its not a replacement and is another option mostly for regression purposes

ry5n:
[...]

PR #1250, opened 2013-04-14
rewrote contrasted utility to not rely on a lightness threshold. Intstea...

bluesaunders:
...d it caluclates brightness of the color and compares to the given background color.

This pull request is for issue #1051.

### DIFF
--- a/doc-src/Gemfile.lock
+++ b/doc-src/Gemfile.lock
@@ -9,28 +9,28 @@ GIT
 
 GIT
   remote: https://github.com/ddfreyne/nanoc.git
-  revision: a326271e25e195dcb381d98f1bf392e2667ec5ec
+  revision: 09f973f67b4768b1d32151289aede81d43dffdd4
   specs:
-    nanoc (3.5.0)
-      cri (~> 2.2)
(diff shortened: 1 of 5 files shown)


mirisuzanne:
Thanks!

chriseppstein:
This is a good patch but it breaks backwards compatibility :(

mirisuzanne:
Should I give these mixins new names, and bring back the old ones with a deprecation warning? I think it's a change that really should be made, and the question is just upgrade path.

chriseppstein:
Ideally, we would have the function generate deprecation messages if the old arguments are passed. And in the case where they are using the old parameters, we would leave the output unchanged. They could silence the deprecation warnings by updating their stylesheets to use the new API.

We should also document this change in the [CHANGELOG](https://github.com/bluesaunders/compass/blob/stable/doc-src/content/CHANGELOG.markdown) and provide the rationale for the change and examples of how to update existing code to use the new API.

mirisuzanne:
Sounds good. @bluesaunders feel free to work on that if you are interested, otherwise I'll try to get to it soon.

PR #1252, opened 2013-04-14
Add --output-dir option to output generated css files to a different directory

zeripath:
Hi,

When using compass as part of a build system, eg. in grunt or yeoman, it is often preferable to generate files in a temporary place, but yet generate them as if they had been created in the source. This keeps the source directory clean. The option --output-dir gives compass the ability to create its css and reference any assets as if the css was being created in --css-dir, but instead of placing the file there, place it in output-dir.

### DIFF
--- a/features/command_line.feature
+++ b/features/command_line.feature
@@ -18,17 +18,17 @@ Feature: Command Line
     And I am told how to conditionally link "IE" to /stylesheets/ie.css for media "screen, projection"
 
   Scenario: Install a project without a framework and separate output-dir
-    Given I should clean up the directory: my_project_output
-    When I create a project using: compass create my_project --output-dir my_project_output
+    Given I should clean up the directory: out
+    When I create a project using: compass create my_project --output-dir out
     Then a directory my_project/ is created
-    And a directory my_project_output/ is created
+    And a directory out/ is created
     And a configuration file my_project/config.rb is created
     And a sass file my_project/sass/screen.scss is created
     And a sass file my_project/sass/print.scss is created
     And a sass file my_project/sass/ie.scss is created
-    And a css file my_project_output/screen.css is created
-    And a css file my_project_output/print.css is created
-    And a css file my_project_output/ie.css is created
+    And a css file out/stylesheets/screen.css is created
+    And a css file out/stylesheets/print.css is created
+    And a css file out/stylesheets/ie.css is created
     And I am told how to link to /stylesheets/screen.css for media "screen, projection"
     And I am told how to link to /stylesheets/print.css for media "print"
     And I am told how to conditionally link "IE" to /stylesheets/ie.css for media "screen, projection"
@@ -42,14 +42,14 @@ Feature: Command Line
     And a css file custom_project/css/screen.css is created
 
   Scenario: Install a project with specific directories and separate output dir
-    Given I should clean up the directory: custom_project_css
-    When I create a project using: compass create custom_project --using compass --sass-dir sass --css-dir css --images-dir assets/imgs --output-dir custom_project_css
+    Given I should clean up the directory: custom_out
+    When I create a project using: compass create custom_project --using compass --sass-dir sass --css-dir css --images-dir assets/imgs --output-dir custom_out
     Then a directory custom_project/ is created
-    And a directory custom_project_css/ is created
+    And a directory custom_out/ is created
     And a directory custom_project/sass/ is created
     And a directory custom_project/css/ is created
     And a sass file custom_project/sass/screen.scss is created
-    And a css file custom_project_css/screen.css is created
+    And a css file custom_out/css/screen.css is created
 
   Scenario: Perform a dry run of creating a project
     When I create a project using: compass create my_project --dry-run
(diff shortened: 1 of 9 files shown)


scottdavis:
I'm not going to merge this without tests I have a feeling its going to cause some regressions

zeripath:
Heh, thought that you might say that.

Didn't write any tests as I'm not sure how to write the tests - in that I don't understand the framework. I know that when it isn't being used there are no regressions, otherwise the current tests would fail - and it certainly works when I run testing manually myself.

I'll have another look at writing some tests, but do you have any hints/pointers for information about the testing framework?

Many thanks for your consideration of the patch though.

andy

scottdavis:
It just uses the standard ruby test unit.

zeripath:
Ah... thanks for forcing me to do this, I've spotted that I wasn't handling the sprites correctly.

There's now some tests in the features section. I hope those are the sort of tests you're looking for.

robwierzbowski:
So if I'm reading this right, you have a directory structure like this:

```
|- css // css_dir
|- sass // sass_dir
|- temp  // output_dir
```

...and identical compiled files are output into the `css` and `temp` directories? Can you explain how this helps with your Yeoman workflow? Not questioning the pull request, just curious because I'm working with Compass and Yeoman too.

zeripath:
Nope that's not the use. Say within dist/ you want to have a directory
structure like:

|- css
|- images |- generated

But within app/ you have:

|- sass
|- images

:; compass compile --app-dir app --css-dir css --generated-images-dir
images/generated --relative-assets #will create within app/:

|- css
|- sass
|- images |- generated

and the css for the generated images will be of the kind:

> url(../images/generated/image.png)

However you want to avoid compass from placing those files within the app
directory, and you don't want to simply just copy over the whole app
directory to dist.  Using the --output-dir option you can simply put:

:; compass compile --app-dir app --css-dir css --generated-images-dir
images/generated --relative-assets --output-dir .tmp

and compass won't place any new files in app/. (Or you could use
--output-dir dist if you like.)

Does that make any more sense?

andy

On 15 April 2013 23:49, Rob Wierzbowski notifications@github.com wrote:

> So if I'm reading this right, you have a directory structure like this:
> 
> |- css // css_dir
> |- sass // sass_dir
> |- temp  // output_dir
> 
> ...and identical compiled files are output into the css and tempdirectories? Can you explain how this helps with your Yeoman workflow? Not
> questioning the pull request, just curious because I'm working with Compass
> and Yeoman too.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/chriseppstein/compass/pull/1252#issuecomment-16416126
> .

## 

Andrew Thornton
art27@cantab.net

robwierzbowski:
That makes more sense. I am guessing that the compass grunt task for distribution gets `--output-dir dist`.

Right now I'm compiling within the app folder and have all generated file directories ignored by git, and moved by a test/build process if necessary.

zeripath:
You can also put --output-dir ../.tmp

The other benefit of this approach is that not all images have to end up
being served under http_images_path or generated_path. So say you had a
stylesheet reference an image in  components/. Just use relative assets and
it falls out. (Although that might not be best practice...)

Andy

zeripath:
Hi Scott,

I was just wondering if there were any more comments on this pull request. I've added some tests now -- used the features test style as that appeared to be the correct way to do it. Would you like me to squash the commits into a single commit?

Andy

scottdavis:
@chriseppstein

chriseppstein:
I don't understand why we need this. you can just set `--css-dir` for a compile and it will go to that new directory for a single compilation.

zeripath:
Say you want to use compass in a build system and you want to keep the
source directory clean; because a) it makes it easier to clean up from a
build and b) it makes ensuring that you don't distribute your sources with
the finished product easier. So you change your css-dir from say: styles to
`../.tmp/styles`

But then you realise that all your image paths are generated '/images/'.
However, your webapp isn't actually meant to be on the root of the server,
so you try to use relative-assets. Now that doesn't work because you end up
with image-urls like `../app/images`. So you end up setting
http_images_path to `../images` and end-up getting rid of relative-assets.

Then you realise that you actually want to use some sass/css from a
components directory. You're not sure which images this refers to, but
really you'd just like the build system to pull those across if they're
referred to. Now most of the stuff to do that relies on parsing the css,
but the urls in the generated css now don't point to those images, they
point to ones in `../images`.

Then of course at some point you want to make a sprite, which means that
you need to set the `http_generated_images_path` and `generated_images_path`.
Then you want to use a font, which also needs the `http_fonts_path` setting
aswell as the font-dir.

Whereas if you are able to set the output-dir to another place then you can
just use relative paths and it all falls out. The images are created in the
output-dir in their own directory, as are the css. image-urls are generated
with reference to where they are from the css-dir. There's no need to have
to set two different settings for the fonts, or the generated-images, or
the images. You can clearly see from the generated css what files from the
components are being referenced - and you can run minimize on them, copy
them to the dist either with the same relative url or rewrite the url.
Meanwhile your source is free from the generated files, and cleaning the
build is quite simple: just delete the output-dir. (I do know that compass
has its own clean method.)

Does it make sense? The end result of this would mean that using compass in
a build system that does not put generated files in the source directory
cleaner and less complicated by gotchas.

chriseppstein:
Ok. I understand. But why does this need to be a compass feature when
simply moving a directory after compilation would accomplish tha same?

zeripath:
Ah... but the whole point of this patch is to prevent compass from creating files from within the source directory, but to allow it to create them as if they are in the source directory. This patch enables compass to do true "out-of-source builds" a) allowing one to build multiple variants (e.g. debug, test, dist) b) prevents accidental committal of generated files c) makes cleaning much easier post build if you use multiple tools. 

(I've just noticed that github has completely messed up my previous post's formatting. Sorry about that, I hope it was more understandable when emailed. Please ask me to repost if it was sent to you completely unformatted.)

chriseppstein:
I guess I don't understand your downstream processes for how such files are put into use. I don't really see the utility over moving the directory after running a compile.

chriseppstein:
You understand why I'm reticent to introduce two kinds of output directories, right? I think this will be very hard for users to understand and will create severe support issues for us.

robwierzbowski:
You can always remove the files with a Yeoman clean command, or just add them to your .gitignore. I think this might be overengineering for a very specific problem.

Issue #1053, opened 2012-10-17
Command line options cannot be extended with custom commands


JohnRiv:
I've developed a custom Compass command line option to be able to run CSS Lint against a project's generated CSS files (see https://github.com/Comcast/compass-csslint), but requiring it in a project's config.rb file is not sufficient enough for the command to become available. The only way I've been able to make it work is to modify /lib/compass/commands.rb and include a require statement there for the command directly.

Example Project:
https://github.com/Comcast/compass-csslint-sample
Note that it includes `require 'compass-csslint'` in the config.rb, but that alone is not enough. You need to modify Compass:

Update to Compass that works:
https://github.com/Comcast/compass/commit/c829eab33bb4087e3e3c8f72385c9e8105dc6702
Note that the only change is including `require 'compass-csslint'` in /lib/compass/commands.rb

With that change to Compass and the compas-csslint gem installed from the repo, you should be able to run `compass csslint` from the command line in the root of a Compass project and see it work

A couple other things I noticed when trying to debug this myself:

If you checkout both compass-csslint-sample & compass-csslint, build the compass-csslint gem locally (`gem build compass-csslint.gemspec`), install the gem in a fresh rvm/gemset for the compass-csslint-sample project (`gem install ../compass-csslint/compass-csslint-0.0.1.gem`), and then run `compass csslint` from the sample project, it returns `Command not found: csslint`

I tried running `compass install csslint` from that project, but that returned `No such framework: "csslint"`, which makes sense since it's a new command & not a framwork.

If you modify lib/compass/commands/registry.rb in Compass to add:
`puts command_class.name`
to the end of def register (line 6) and installed that modified version, then run `compass csslint` from the sample project, the last line it outputs is:
`Compass::Commands::UnpackExtension`
Interestingly, if I run "compass install csslint", the last line it outputs is
`Compass::Commands::CSSLintProject`
But it's still followed by `No such framework: "csslint"`

paulirish:
John's work here with csslint for compass is fantastic. Can't wait to see much easier [installation instructions](http://comcast.github.com/compass-csslint/) :)  :+1:

MoOx:
:+1:

aglemann:
+1

scottdavis:
Since this is an api that needs to be created do you have any input on how you would like to hook in to compass to register a command?

matt-d-rat:
+1 for this to be fixed and included in the next version of compass, would be super useful!

[...]

Issue #1200, opened 2013-03-05
Ad-hoc extension does not work


hlb:
The "Ad-hoc extension" feature does not work in both Compass 0.12.2 and 0.13.alpha 2 http://compass-style.org/help/tutorials/extensions/

https://gist.github.com/hlb/5087139

Or is there anything I have missed?

scottdavis:
will get this fixed for 0.13

robwierzbowski:
+1. Ad hoc extensions are a perfect solution for front-end package managers like Bower. Right now we're limited to using Sass components, but when this is fixed we'll have access to Compass extensions as components for app builds.

This works for me in 0.12.2, but I've heard it doesn't in 0.13.alpha.4.

cimmanon:
I just updated Compass to 0.13.alpha.4 (Markab) and now one of my ad hoc extensions is ignored and it appears to be because of the name of the directory.

```
// doesn't work
extensions/cimma
// works
extensions/cimm
extensions/cimma-bak
extensions/cimma-
extensions/cimmaa
```

Error from the console:

```
>>> Compass is watching for changes. Press Ctrl-C to Stop.
    error sass/test.scss (Line 1: File to import not found or unreadable: cimma.
Load paths:
  /home/chris/compass-extensions/sass
  /usr/local/lib/ruby/gems/1.9.1/gems/compass-0.13.alpha.4/frameworks/compass/stylesheets
  /home/chris/compass-extensions/extensions/foo/stylesheets
  /home/chris/compass-extensions/extensions/gmedia/stylesheets
  /home/chris/compass-extensions/extensions/normalize/stylesheets
  Compass::SpriteImporter)
```

My load paths listing doesn't even include `/home/chris/compass-extensions/extensions/cimma/stylesheets`, but all of my other ad hoc extensions are listed.  This worked without issue in previous versions of Compass.

cimmanon:
I found the problem.  I had a new extension where I missed renaming it in lib/_my-extension_.rb.

chriseppstein:
This should be fixed in 1.0.0 please test against the alphas.

PR #1261, opened 2013-04-19
Adding ability to extend command line options with custom extensions

JohnRiv:
Here is my proposal to fix issue #1053

This will allow anyone to include custom command line extension for Compass. In order for a command line extension to work, it must be available as a gem that is prefixed with "compass-". Users can then include that in their projects by adding a command_extensions property to their config.rb file, such as:

``` ruby
command_extensions = ["csslint"]
```

The above example would include the [compass-csslint](https://github.com/Comcast/compass-csslint) command line extension.

command_extensions is intended to take an array of strings, although I did also add support for it to be just a string if only 1 extension is desired to be included. The strings can include the "compass-" prefix as well:

``` ruby
command_extensions = ["compass-csslint"]
```

### DIFF
--- a/features/command_line.feature
+++ b/features/command_line.feature
@@ -183,7 +183,7 @@ Feature: Command Line
     Given ruby supports fork
     Given I am using the existing project in test/fixtures/stylesheets/compass
     When I run: compass compile
-    And I run in a separate process: compass watch 
+    And I run in a separate process: compass watch
     And I wait 3 seconds
     And I touch sass/layout.sass
     And I wait 2 seconds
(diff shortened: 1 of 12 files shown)


scottdavis:
Looks good. Except it probably doesn't work with adhoc extensions

JohnRiv:
My follow up commit makes it work for ad-hoc extensions, and also appears to fix issue #1200 as well.

JohnRiv:
Any feedback on my latest commit? If we can get this merged in, I'd like to talk about this feature during the 2nd day of SassConf as one of the unconference sessions to help spread the word.

chriseppstein:
This landed on master last week.

PR #1266, opened 2013-04-22
Abstract digest class

agelwarg:
When using compass to compile sprites, we have run into an issue where Digest::MD5 is not an acceptable digest algorithm -- specifically for (Linux) hosts running in FIPS-enabled mode.  As such, this proposed change is what we're using to abstract the digest class used.  The default is Digest::MD5, but in our application we override it with Digest::SHA2 in our compass.config.

I'm not sure what tests I can/should create.

### DIFF
--- a/lib/compass/configuration.rb
+++ b/lib/compass/configuration.rb
@@ -40,6 +40,7 @@ module Compass
       :preferred_syntax,
       :disable_warnings,
       :sprite_engine,
+      :digest_class,
       :chunky_png_options
     ].flatten
 
--- a/lib/compass/configuration/defaults.rb
+++ b/lib/compass/configuration/defaults.rb
@@ -156,6 +156,10 @@ module Compass
         :chunky_png
       end
       
+      def default_digest_class
+        Digest::MD5
+      end
+
       def default_chunky_png_options
         {:compression => Zlib::BEST_COMPRESSION}
       end
--- a/lib/compass/sass_extensions/sprites/image.rb
+++ b/lib/compass/sass_extensions/sprites/image.rb
@@ -97,9 +97,9 @@ module Compass
           @spacing ||= (get_var_file("spacing") || options.get_var("spacing") || Sass::Script::Number.new(0, ['px'])).value
         end
 
-        # MD5 hash of this file
+        # Digest/hash of this file
         def digest
-          Digest::MD5.file(file).hexdigest
+          Compass.configuration.digest_class.file(file).hexdigest
         end
         
         # mtime of this file
--- a/lib/compass/sass_extensions/sprites/sprite_methods.rb
+++ b/lib/compass/sass_extensions/sprites/sprite_methods.rb
@@ -78,7 +78,7 @@ module Compass
         # Returns the uniqueness hash for this sprite object
         def uniqueness_hash
           @uniqueness_hash ||= begin
-            sum = Digest::MD5.new
+            sum = Compass.configuration.digest_class.new
             sum << SPRITE_VERSION
             sum << path
             sum << layout


scottdavis:
Out of curiosity why can't md5 be used with ruby on fips it comes standard when you build ruby

agelwarg:
From what i understand, MD5 is simply not a FIPS-approved hashing
algorithm.  You can search the net and you'll see all kinds of references
in many different applications across all platforms.  Some solutions are
just replacing the hard coding (which isn't really good for the long term
as the new algorithm could become non-compliant as well down the line,
requiring another change) while others are making it configurable, such as
my proposal.

scottdavis:
Sure, but you are not using the hash for a security purpose so why does it matter i guess was my real question.

agelwarg:
I completely agree with you, but what you will find is that if you run your
OS in FIPS-enabled mode, using MD5 (apparently for any purpose) will fail.
What I have not been able to figure out though is exactly which MD5
implementation causes the failure.  I know that your code simply references
Digest::MD5, but it must be overridden somewhere in the stack.   For
example, this is from a FIPS-enabled OS (RHEL6) with REE
(ruby-enterprise-1.8.7-2012.02):

irb(main):001:0> require 'digest/md5'
=> true
irb(main):002:0> Digest::MD5.hexdigest("hello")
=> "5d41402abc4b2a76b9719d911017c592"

irb(main):003:0> require 'openssl'
=> true
irb(main):004:0> OpenSSL::Digest::MD5.hexdigest("hello")
digest.c(149): OpenSSL internal error, assertion failed: Digest init
previous FIPS forbidden algorithm error ignored
Aborted (core dumped)

scottdavis:
Do me a favor and try to replicate this in your irb terminal and let me know what you get: https://github.com/chriseppstein/compass/blob/stable/lib/compass/sass_extensions/sprites/sprite_methods.rb#L86-L95 I
 may be able to clean this up a bit to use `Digest::MD5.hexdigest` since that seems to work.

scottdavis:
Just to clarify the reason I'm hesitant to go with a SHA hash is compile speed. I wouldn't have a problem if it was faster just switching to a SHA based hash. Really in a perfect world crc16 would be best for creating this hash if it wasn't so collusion prone.

As for creating a new config option of this. IT seems over kill if for your edge case problem overriding Digest:MD5 = Digest:SHA works then that just may have to be your solution.

agelwarg:
What do you mean replicate it -- do you have a small script to run that
explicitly triggers that part of the code?

scottdavis:
basically im curious why `require 'digest/md5' => true irb(main):002:0> Digest::MD5.hexdigest("hello") => "5d41402abc4b2a76b9719d911017c592"` worked and our implementation does not.

agelwarg:
Let me build a better irb example to demonstrate what I'm seeing happen --
I'll be in touch as soon as I can.

chriseppstein:
This is an implementation detail and I don't want to expose it to our public API.

If the MD5 class doesn't work in certain modes, we should detect that mode and select a digest that works.

Can we rework this patch to select a digest class automagically? Closing this as-is.

voxik:
Hi,

Could you please reconsider this? MD5 is disabled in OpenSSL by default on RHEL7 and on recent versions of Fedora.

@agelwarg Digest::MD5 is typically implemented by OpenSSL::Digest::MD5 but I suspect, that when you compile Ruby on FIPS enabled machine, there kicks in default implementation of MD5, which is independent of OpenSSL. Actually, if you disable OpenSSL extension during build, you still should be able to use Digest::MD5.hexdigest (but I have not tested it, so I might be wrong).

chriseppstein:
I'm not going to expose a public API for this. It's an implementation detail. If someone will file a bug with reproducible steps where compass sprites don't work, we will create a fix.

scottdavis:
@voxik can you post a link to an iso of a FIPS distro I want to do some poking around.

voxik:
I am not aware about any distro with FIPS enabled by default, but this [1] is guide how to enable FIPS for RHEL/CentOS 6 and this [2] is for RHEL/CentOS 7. I assume that similar steps applies for Fedora.

[1] https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Security_Guide/sect-Security_Guide-Federal_Standards_And_Regulations-Federal_Information_Processing_Standard.html

[2] https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Security_Guide/chap-Federal_Standards_and_Regulations.html#sec-Enabling-FIPS-Mode

PR #1273, opened 2013-04-26
Flexbox mixin adjustments

cimmanon:
Added a variable here to suppress `box-direction: normal` in instances where it is unnecessary.  Not sure if this should default to true or false, so I've set it to true so that the test doesn't have to be updated.

``` scss
$flex-legacy-enabled: true;
.foo {
    @include flex-direction(row);
}
```

Output:

``` css
.foo {
  -webkit-box-orient: horizontal;
  -moz-box-orient: horizontal;
  -webkit-box-direction: normal;
  -moz-box-direction: normal;
  -webkit-flex-direction: row;
  -ms-flex-direction: row;
  flex-direction: row;
}
```

Since `normal` is the default value for `flex-direction`, there's no reason to to emit it unless we're actually overriding a reverse value set elsewhere.

``` scss
$flex-legacy-enabled: true;
.foo {
    @include flex-direction(row, $normal: false);
}
```

Output:

``` css
.foo {
  -webkit-box-orient: horizontal;
  -moz-box-orient: horizontal;
  -webkit-flex-direction: row;
  -ms-flex-direction: row;
  flex-direction: row;
}
```

Unless someone can find a browser that supports the draft from early 2012 under the -webkit` prefix, I'm recommending setting it to false.  The sources I've been able to dig up indicate that all the mobile devices are either 2009 (the Android emulator confirms this is true for the most recent version, can't confirm for iOS) or standard (Blackberry 10).  Chrome does not support it in v26 (would have to do some digging to find out exactly when it was dropped).

Quick support demo:  http://codepen.io/cimmanon/pen/gFvmx

There seem to be a few other minor changes that slipped in that should have been a separate commit (function was renamed and global variables were reorganized).  Sorry about that.

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_flexbox.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_flexbox.scss
@@ -231,25 +231,25 @@ $flex-legacy-orient: true !default;
 
 // Distributing extra space along the "cross axis"
 // $value: flex-start | flex-end | center | space-between | space-around | stretch
-@mixin align-content($value: flex-start, $wrap: $flex-wrap-required, $legacy: $flex-legacy-enabled) {
-	@if $legacy and not $wrap {
-		@include legacy-align-content($value);
-	}
+@mixin align-content($value: flex-start) {
+	// There is no @box version of this property
 
 	@include experimental(flex-line-pack, flex-translate-value($value, flexbox), $flexbox-support...);
 	@include experimental(align-content, $value, $flex-support...);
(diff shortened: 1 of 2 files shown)


chriseppstein:
@cimmanon I've recently made some changes in the flexbox module. Can you rebase this branch against master?

cimmanon:
I had to do it by hand.  With the addition of disabling the webkit prefix via `$flexbox-support`, the test will need to be redone.

PR #1283, opened 2013-05-11
Make sure Date.today is defined in the gemspec.

pwnall:
This was necessary to make compass work when loaded using bundler's :path option.

I'm on MRI 1.9.3.


scottdavis:
its better to just remove the date from the gemspec which has already been done in master.

scottdavis:
https://github.com/chriseppstein/compass/commit/b7e8ba265f1d715b2a541b51bcede3a09c459729

Issue #1289, opened 2013-05-21
Sticky footer doesn't go to bottom of document when page is taller than viewport


daltonrooney:
If I use the sticky footer mixin and my page is taller than 100% of the viewport, the body element ends at the bottom of the viewport, not the bottom of the page. This leaves a gap between my background image and the bottom of the document. Simple fix, set a min-height of 100% on the body and a height: auto.

![before](https://f.cloud.github.com/assets/742045/544201/07d6a4d6-c24b-11e2-9f5a-1735d95c0d02.png)
Before

![after](https://f.cloud.github.com/assets/742045/544202/07dfe49c-c24b-11e2-9aef-dbf0420405e4.png)
After

daltonrooney:
Sorry, I meant to attach the pull request to this issue. https://github.com/chriseppstein/compass/pull/1290

daltonrooney:
Turns out that this was a conflict with the normalize stylesheet included with Zurb Foundation 4. Problem solved.

PR #1290, opened 2013-05-21
#1289 Update _sticky-footer.scss

daltonrooney:
Make sure body extends to bottom of document, not just 100% of page, so background images continue below the footer.


scottdavis:
I would make this change against master and up date it with a test case.

daltonrooney:
Thanks for the advice. I also realized that it needs more testing!

PR #1295, opened 2013-05-23
Css3 browser-support handling

mirisuzanne:


cimmanon:
There's still a naming conflict between the support variables in the box and flexbox files.  The flexbox file uses `$flex-support,`$flexbox-support`, and`$box-support`variables which correspond to the display properties of the same name set as`!default`.  The box file uses`$flexbox-support` as its variable, which overwrites the wrong variable in the flexbox file.

chriseppstein:
@cimmanon I have a fix for that I'm about to push.

chriseppstein:
@cimmanon That patch is here: d89db7cfc759f038dc70523f30cd0a26e5608ec1

PR #1301, opened 2013-05-30
Removed -moz-box-shadow

robinbijlani:
Gecko 13 (Firefox 13) removed support for -moz-box-shadow. Since then, only the unprefixed version is supported. https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow


neilhem:
better write `-webkit, not -moz, not -o, not -ms, not -khtml, official`

Anahkiasen:
I'm not sure `-webkit` is [needed anymore](http://caniuse.com/#search=box-shadow) for that matter.

chriseppstein:
This is rewritten in 1.0 and driven by usage thresholds and caniuse.com data.

PR #1303, opened 2013-05-30
Smart layout now considers spacing, should fix #718

### LINKED ISSUE #718: sprite with smart layout not support sprite-spacing?
When i use sprite-layout with smart, sprite-spacing not work, does it not support sprite-spacing with smart layout?

$icon-spacing: 5px !default;
$icon-layout:"smart";

Raibaz:
I basically added the spacing to every width and height computation, both for every image row in the smart layout fitter and in the fitter itself.

Since the output of this commit is a .png file with the sprite, I don't know how to build tests for this; however, since the changes are quite small and straightforward, and I have run some testing on different sets of images myself, I think it shouldn't break any of the existing behaviors.

Attached is a pair of sprites generated with smart layout, with 30px spacing and without the spacing rule in the .scss:

No spacing:
registrazione-s05deac8370

30px spacing:
registrazione-s9bc14f8201

### DIFF
--- a/lib/compass/sass_extensions/sprites/image_row.rb
+++ b/lib/compass/sass_extensions/sprites/image_row.rb
@@ -9,13 +9,13 @@ module Compass
         attr_reader :images, :max_width
         def_delegators :@images, :last, :delete, :empty?, :length
         
-        def initialize(max_width)
+        def initialize(max_width)          
           @images = []
           @max_width = max_width
         end
         
         def add(image)
-          return false if !will_fit?(image)
+          return false if !will_fit?(image)          
           @images << image
           true
         end
@@ -23,15 +23,15 @@ module Compass
         alias :<< :add
         
         def height
-          images.map(&:height).max
+          return (images.map(&:height).max + images.map(&:spacing).max)
         end
         
-        def width
+        def width          
           images.map(&:width).max
         end
 
         def total_width
-          images.inject(0) {|sum, img| sum + img.width }
+          images.inject(0) {|sum, img| sum + img.width + img.spacing}
         end
         
         def efficiency
@@ -39,7 +39,7 @@ module Compass
         end
 
         def will_fit?(image)
-          (total_width + image.width) <= max_width
+          (total_width + image.width + image.spacing) <= max_width
         end
       end
     end
--- a/lib/compass/sass_extensions/sprites/row_fitter.rb
+++ b/lib/compass/sass_extensions/sprites/row_fitter.rb
@@ -9,7 +9,7 @@ module Compass
         attr_reader :images, :rows
         def_delegators :rows, :[]
 
-        def initialize(images)
+        def initialize(images)          
           @images = images.sort do |a,b|
             if a.height == b.height
               b.width <=> a.width
@@ -17,7 +17,7 @@ module Compass
               a.height <=> b.height
             end
           end
-          @rows = []
+          @rows = []        
         end
 
         def fit!(style = :scan)
@@ -26,11 +26,11 @@ module Compass
         end
 
         def width
-          @width ||= @images.collect(&:width).max
+          @width ||= @images.collect(&:width).max + @images.collect(&:spacing).max
         end
         
         def height
-          @height ||= @rows.inject(0) {|sum, row| sum += row.height}
+          @height ||= @rows.inject(0) {|sum, row| sum += row.height} + @images[0].spacing
         end
 
         def efficiency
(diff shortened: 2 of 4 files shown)


scottdavis:
Unit testing should be do e for the row fitter classes these tests are located in the test/units/sprite folder also I would really prefer this pr be made against theater branch

lib/compass/sass_extensions/sprites/image_row.rb
```diff
@@ -23,15 +23,15 @@ def add(image)
         alias :<< :add
         
         def height
-          images.map(&:height).max
+          return (images.map(&:height).max + images.map(&:spacing).max)
```
chriseppstein, line 26: I don't think that's right. I think it's `images.map {|i| i.height + i.spacing}.max`

Also, we don't need a `return` here. The ruby idiom is to only use `return` for early returns.

lib/compass/sass_extensions/sprites/image_row.rb
```diff
@@ -23,15 +23,15 @@ def add(image)
...
         end
         
-        def width
+        def width          
           images.map(&:width).max
         end
 
         def total_width
-          images.inject(0) {|sum, img| sum + img.width }
+          images.inject(0) {|sum, img| sum + img.width + img.spacing}
```
chriseppstein, line 34: This doesn't reused the space from the previous image. The space between image N and image N+1 should be the max of the spacing around N and N+1.

lib/compass/sass_extensions/sprites/image_row.rb
chriseppstein, line 26: Actually do we need spacing around the edges?

lib/compass/sass_extensions/sprites/image_row.rb
chriseppstein, line 34: But even without space reuse this is wrong because there's space on both sides. The naive algorithm would be `images.inject(0) {|sum, img| sum + img.width + 2 * img.spacing}`. Also, do we need spacing on the edges? I don't think so.

lib/compass/sass_extensions/sprites/image_row.rb
```diff
@@ -39,7 +39,7 @@ def efficiency
         end
 
         def will_fit?(image)
-          (total_width + image.width) <= max_width
+          (total_width + image.width + image.spacing) <= max_width
```
chriseppstein, line 42: If we don't need spacing on the edges this is right unless we get the spacing between adjacent images to be shared. In which case, this would only be the delta between the space of the previous image and this image if this image has more spacing.

lib/compass/sass_extensions/sprites/row_fitter.rb
```diff
@@ -26,11 +26,11 @@ def fit!(style = :scan)
         end
 
         def width
-          @width ||= @images.collect(&:width).max
+          @width ||= @images.collect(&:width).max + @images.collect(&:spacing).max
```
chriseppstein, line 29: Either this is 2x the spacing or it needs to share the spacing properly. Also: the question about edge spacing again.

lib/compass/sass_extensions/sprites/row_fitter.rb
```diff
@@ -26,11 +26,11 @@ def fit!(style = :scan)
...
         end
         
         def height
-          @height ||= @rows.inject(0) {|sum, row| sum += row.height}
+          @height ||= @rows.inject(0) {|sum, row| sum += row.height} + @images[0].spacing
```
chriseppstein, line 33: 2x spacing again. Or maybe we don't need edge spacing at all.

Raibaz:
I guess the main mistake I made is considering just the global sprite spacing and ignoring the notion that every image may have its own spacing, so @chriseppstein's observations are all correct.

Going to fix this.

lib/compass/sass_extensions/sprites/row_fitter.rb
chriseppstein, line 33: I guess the row would need to know if it's the first or last row to be able to calculate this correctly should we not need edge spacing.

chriseppstein:
:+1: for the master branch.

:-1: for all the trailing whitespace that got added.

My review showed a number of logic failures throughout this patch. I'm not sure that "eye-balling it" is working. We need to construct specific test cases, even if the output of those test cases is visually verified.

Raibaz:
Fixing the logic issues you pointed out right now, then going to build a specific test and remove the trailing whitespaces.

estahn:
+1

estahn:
@chriseppstein Was this merged into compass?

TommySorensen:
@chriseppstein Was this merged into compass? I can't se this working in the latest --pre build from compass=

camilonova:
@chriseppstein I can't see this working too, it's already merged?

scottdavis:
No, this hasn't been merged this was made against stable and needs to be rebased to master since the internal api has changed some. And there are some code quality issues.

PR #1308, opened 2013-06-01
Compiler cache

scottdavis:


thedarkone:
Hey,

since when is compass/sass thread safe? Otherwise you should look into `ThreadSafe::Cache` from the [thread_safe](https://github.com/headius/thread_safe) gem (used internally by Rails 4.0).

lib/compass/compiler/cache.rb
```diff
@@ -0,0 +1,88 @@
...
+      end
+
+      semaphore.synchronize do
+        @hash[key] = value
+      end
+    end
+
+    def clear!
+      semaphore.synchronize do
+        @hash = nil
```
thedarkone: This is unnecessary.

lib/compass/compiler/cache.rb
```diff
@@ -0,0 +1,88 @@
...
+    # ==== Examples
+    #
+    # Basic Usage with block
+    #     foo = @cache.cache('foo') { 15 }
+    #     foo will return 15
+    # Basic Usage without block
+    #     foo = @cache.cache('foo', 15)
+    #     foo will return 15
+    def cache(key, value=nil, &block)
+      return read(key) if has_key?(key) #return the object if it already in the cache
```
thedarkone: This might return incorrect values (`nil`s for example) under concurrent load.

lib/compass/compiler/cache.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    extend Forwardable
+    attr_reader :hash, :semaphore
+
+    # use the delgator pattern to pass through methods to hash
+    # I didn't extend hash because we have a need to beable to clear and re initalize the hash
+    # and you can not reassign self within a class instance
+
+    def_delegator :@hash, :delete, :invalidate!
+    def_delegator :@hash, :[], :read
```
thedarkone: Technically read access also needs to be synchronized.

lib/compass/compiler/cache.rb
```diff
@@ -0,0 +1,88 @@
...
+
+
+    extend Forwardable
+    attr_reader :hash, :semaphore
+
+    # use the delgator pattern to pass through methods to hash
+    # I didn't extend hash because we have a need to beable to clear and re initalize the hash
+    # and you can not reassign self within a class instance
+
+    def_delegator :@hash, :delete, :invalidate!
```
thedarkone: Needs to be synchronized.

scottdavis:
The new compiler will be thread safe. 

Also that thread_safe gem looks awesome .. I'm going to rip all this out and use that. nice work

scottdavis:
Btw also want to say thanks for the input my thread safe programing in ruby is really rusty

thedarkone:
`ThreadSafe::Cache` should work on all Rubies, [see here](https://github.com/headius/thread_safe/blob/master/lib/thread_safe/cache.rb#L10-L18).

thedarkone:
Also, sorry for no docs, if you decide to use it, you basically care about that it behaves just like a `Hash` and has a `compute_if_absent(key, &block)` method.

scottdavis:
I think I'm just going to wrap my cache function around put_if_absent and call it a day

compass.gemspec
```diff
@@ -16,7 +16,8 @@ Gem::Specification.new do |gemspec|
   gemspec.add_dependency 'sass', '>= 3.2.5', '< 3.4'
   gemspec.add_dependency 'chunky_png', '~> 1.2'
   gemspec.add_dependency 'listen', '~> 1.0.0'
-
+  gemspec.add_dependency "thread_safe", "~> 0.1.0"
```
chriseppstein, line 19: I don't want to add this dependency. Let's just use mutexes.

chriseppstein, line 19: Reason: I had some serious performance issues with it at work and I just don't think a complex dependency is necessary here because a simple mutex around reading or writing to the cache is good enough.

scottdavis, line 19: That's what I originally had and everyone freaked out!

chriseppstein, line 19: You were right.

irunatbullets:
Any ETA on this being pulled into the main branch?

scottdavis:
Still have some work todo on this chris and i were talking and we arnt
going to use the thread_safe gem

mmehlhope:
Looking forward to seeing this pulled in. This bug is our team's single largest performance hit in regards to static asset compilation -- thanks for all the work to get it fixed.

chriseppstein:

Issue #1244, opened 2013-04-03
Allow Sprite engines to define allowed image formats


scottdavis:
via #292

kurtdavis:
Could we add this [comment](https://github.com/chriseppstein/compass/pull/1258) to the docs while sprite engines don't support other formats than PNG?  Then expand documentation as needed later.
There has been many failed attempts by early adopters due to the cryptic nature of the error message when trying to import '*.gif' images.

Issue #907, opened 2012-05-28
JPG sprites support


josefrichter:
Please, would it be possible to implement support for JPG sprites, too? Sprites are being used much more widely these days than just for small PNG icons. But I don't have the skills to implement it.

chriseppstein:
The image manipulation library we use is pure-ruby which makes it very easy for us to support it on all platforms. However, it's possible to support this use case if you use https://github.com/Compass/compass-rmagick-engine or if we can find a pure-ruby implementation of jpeg.

scottdavis:
the rmagick engine only does pngs also its just a bit faster ... the importer it self verifys that all images are png's and alot of png assumptions are taken within the spirite image class methods. TL;DR version this will take some rewriting to impliment

scottdavis:
also if anyone wants to write a jpeg library in pure ruby alot of people would be excited

robwierzbowski:
With jpegs there's also a background color issue — no transparency, so the element containing the sprite has to be exactly the correct size, or a background color needs to be passed into the sprite generator. This is one of the reasons sprites are generally pngs only, and IMO one of the good reasons to keep them so.

chriseppstein:
I agree with @robwierzbowski, JPEG sprites are not a sprite-friendly image format. However, SVG and GIF are viable.

beatak:
this may be a bit out of context, but @scottdavis I saw https://github.com/chriseppstein/compass/commit/520c950f7d9a9bf7d5cf99842e898feec8aa7364 and it seems this will let me order the sprite file by file name or file size?  that looks very very interesting and I wish it'll be merged to master!

scottdavis:
520c950 is on master

PR #1312, opened 2013-06-05
Allowed extensions

scottdavis:
Allows sprite engines to define allowed extensions

### DIFF
--- a/lib/compass/sass_extensions/sprites/engines.rb
+++ b/lib/compass/sass_extensions/sprites/engines.rb
@@ -1,9 +1,10 @@
+
 module Compass
   module SassExtensions
     module Sprites
       class Engine
         attr_accessor :width, :height, :images, :canvas
-        def initialize(width, height, images)
+        def initialize(width=nil, height=nil, images=nil)
           @width, @height, @images = width, height, images
           @canvas = nil
         end
@@ -20,6 +21,5 @@ module Compass
     end
   end
 end
-
 
 require 'compass/sass_extensions/sprites/engines/chunky_png_engine'
--- a/lib/compass/sass_extensions/sprites/sprite_map.rb
+++ b/lib/compass/sass_extensions/sprites/sprite_map.rb
@@ -32,6 +32,10 @@ module Compass
           end
         end
 
+        def self.sprite_engine_class      
+          constantize("Compass::SassExtensions::Sprites::#{self.modulize}Engine")
+        end
+
         def initialize(sprites, path, name, context, kwargs)
           @image_names = sprites
           @path = path
@@ -81,9 +85,22 @@ module Compass
 
         private 
 
-        def modulize
+        def self.modulize
           @modulize ||= Compass::configuration.sprite_engine.to_s.scan(/([^_.]+)/).flatten.map {|chunk| "#{chunk[0].chr.upcase}#{chunk[1..-1]}" }.join
         end
+
+        def self.constantize(camel_cased_word)
+          names = camel_cased_word.split('::')
+          names.shift if names.empty? || names.first.empty?
+
+          constant = Object
+          names.each do |name|
+            constant = constant.const_defined?(name) ? constant.const_get(name) : constant.const_missing(name)
+          end
+          constant
+        end
+
+
 
       end
     end
--- a/lib/compass/sass_extensions/sprites/sprite_methods.rb
+++ b/lib/compass/sass_extensions/sprites/sprite_methods.rb
@@ -18,7 +18,7 @@ module Compass
         end
         
         def init_engine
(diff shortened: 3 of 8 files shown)


scottdavis:
This should allow #907

scottdavis:
fixes #1244

lib/compass/sass_extensions/sprites/engines.rb
```diff
@@ -1,9 +1,10 @@
+
 module Compass
   module SassExtensions
     module Sprites
       class Engine
         attr_accessor :width, :height, :images, :canvas
-        def initialize(width, height, images)
+        def initialize(width=nil, height=nil, images=nil)
```
chriseppstein, line 7: what is this change for?

lib/compass/sass_extensions/sprites/sprite_map.rb
```diff
@@ -81,10 +85,23 @@ def method_missing(meth, *args, &block)
...
+
+          constant = Object
+          names.each do |name|
+            constant = constant.const_defined?(name) ? constant.const_get(name) : constant.const_missing(name)
+          end
+          constant
+        end
+
+
+
```
chriseppstein, line 104: neeeeeewlineeeeees.

lib/compass/sprite_importer.rb
```diff
@@ -1,10 +1,14 @@
 require 'erb'
+require 'compass'
 require 'compass/sprite_importer/binding'
+
+
+
+
```
chriseppstein, line 7: neeeeeewlineeeeees!

lib/compass/sass_extensions/sprites/sprite_methods.rb
```diff
@@ -18,7 +18,7 @@ def compute_image_metadata!
         end
         
         def init_engine
-          @engine = eval("::Compass::SassExtensions::Sprites::#{modulize}Engine.new(nil, nil, nil)")
+          @engine = self.class.sprite_engine_class.new
           @engine.width = @width
           @engine.height = @height
           @engine.images = @images
```
chriseppstein, line 24: Why not just pass these arguments to the initializer instead of making them optional?

lib/compass/sprite_importer.rb
```diff
@@ -1,10 +1,14 @@
...
 module Compass
   class SpriteImporter < Sass::Importers::Base
     VAILD_FILE_NAME       = /\A#{Sass::SCSS::RX::IDENT}\Z/
-    SPRITE_IMPORTER_REGEX = %r{((.+/)?([^\*.]+))/(.+?)\.png}
-    VALID_EXTENSIONS      = ['.png']
+    SPRITE_IMPORTER_REGEX = %r{((.+/)?([^\*.]+))/(.+?)}
```
chriseppstein, line 11: I'm ok with relaxing this glob restriction, but unless the user specifies a valid extension in the glob explicitly, I think we should return `nil` from find unless we find an image with a valid extension in the folder. This will keep it from conflicting with globs that import sass.

scottdavis, line 11: So the entire point of this was to move the required extensions down a level to the engine classes so things like svg gif or whatever people wanted could be created and also allow me to move forward using the same code to work on a svg -> font glyph engine

lib/compass/sprite_importer.rb
```diff
@@ -107,6 +111,16 @@ def self.content_for_images(uri, name, skip_overrides = false)
...
+
+  private
+
+    def self.valid_extensions
+      @valid_extensions ||= SassExtensions::Sprites::SpriteMap.sprite_engine_class::VALID_EXTENSIONS
+    end
+
+    def self.sprite_importer_regex_with_ext
+      @importer_regex ||= %r{#{SPRITE_IMPORTER_REGEX}(#{valid_extensions.join('|')})}
+    end
```
scottdavis, line 123: So here is where the glob is scoped.

lib/compass/sass_extensions/sprites/engines.rb
scottdavis, line 7: no idea was 4 months ago :(

lib/compass/sprite_importer.rb
chriseppstein, line 11: That's fine. My point is that when the glob matches, we always return a sass engine. Instead, the sprite engine should be able to look at that glob and decide if it has files in it that it can process. If not, this importer should return nil and then a different importer can handle it. Or maybe Sass will just give an error that says that it cannot import that.

lib/compass/sprite_importer.rb
chriseppstein, line 123: I see. I'd still like to not return any engine if there's no files that match the glob.

chriseppstein:

chriseppstein:
Should be on top of the changes coming in #1771.

PR #1316, opened 2013-06-11
Update _flexbox.scss

timhettler:
Notes:
- I don't think browser support is in a place where `$flex-legacy-enabled` should be false by default. It gets situationally set to false if you're doing advanced flexbox stuff anyway, so an "opt-out" policy seems better overall.
- Support for `align-content` & `align-items` was mixed up. `align-content` only takes effect when there's line wrapping, so legacy support is out. `align-items` (`box-align` in the legacy syntax) is the property responsible for cross-axis alignment. (See: http://timhettler.github.io/compass-flexbox/tests.html#align-items)

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_flexbox.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_flexbox.scss
@@ -53,7 +53,7 @@ $box-support: -moz, -webkit, not -o, not -ms, not -khtml, not standard !default;
 
 // If `true`, the @box properties will be emitted as part of the normal mixin call
 // (if this is false, you're still free to explicitly call the legacy mixins yourself)
-$flex-legacy-enabled: false !default;
+$flex-legacy-enabled: true !default;
 
 // @doc off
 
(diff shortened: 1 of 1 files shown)


cimmanon:
`$flex-legacy-enabled` is turned off by default as recommended by Eric Meyer (see: https://github.com/chriseppstein/compass/issues/921#issuecomment-13022147).  This isn't just a case of "match the right properties up and everything works" or "pretty much every browser supports the new implementation, so we're dropping the old stuff".  There are some serious differences between the 2009 implementations and the modern ones.  "I need flex items to wrap" is not the only reason to turn off legacy support (the `$flex-wrap-required` variable does more than just suppress the old 2009 properties from being emitted automatically).

These are the most significant scenarios I've run across that aren't bug related:
- http://stackoverflow.com/questions/16130399/flexbox-in-safari-grow-and-shrink
- http://stackoverflow.com/questions/16122341/crucial-difference-between-flexbox-2009-and-2012-in-chrome

I think there's 5 serious bugs with Firefox's 2009 Flexbox (thankfully they're moving to the standard version, but mobile versions will linger forever).  The Webkit 2009 implementation has issues with floated flex items (floating should be ignored).

I have no objections with it being enabled by default.  I do feel that it is safer to turn it off by default.  This wouldn't be the first instance of a feature disabled by default (see SVG gradient fallback).

chriseppstein:
The flexbox module was rewritten for 1.0 and no longer tries to create a unified mixin that spans these divergent specs.

PR #1323, opened 2013-06-22
Improved cache buster handling for font URLs

thomasbachem:
This is an improvement/addition to 727ac80a677af21856de55f575b54387b7b53503.

Font URLs do often contain query strings and/or hashes (e.g. "example.eot?#iefix" or "example.svg#my-font"). This patch adds support for cache busters in combination with such font URLs.

Besides that, I changed the default value of font_url's :cache_buster argument to true, just like with image_url. I think that's more consistent.

### DIFF
--- a/lib/compass/sass_extensions/functions/urls.rb
+++ b/lib/compass/sass_extensions/functions/urls.rb
@@ -47,7 +47,7 @@ module Compass::SassExtensions::Functions::Urls
         base.declare :font_url,       [:path, :only_path, :cache_buster]
       end
     end
-    def font_url(path, only_path = Sass::Script::Bool.new(false), cache_buster = Sass::Script::Bool.new(false))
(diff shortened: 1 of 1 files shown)


chriseppstein:
I like where this is going but we need tests for this.

chriseppstein:
I implemented this myself, thanks for the patch. http://git.io/H-myBA

thomasbachem:
Thanks for implementing it!

thomasbachem:
@chriseppstein Your solution doesn't handle query strings, especially the typical `example.eot?#iefix` syntax. The `file` argument to the cachebuster will be empty in this case, as he looks for `example.eot?` instead of `example.eot`. In addition, returning any cachebuster will lead to two question marks, e.g. `example.eot??CACHEBUSTER#iefix`.

PR #1326, opened 2013-07-01
Multiple Output Styles 

Ne-Ne:
Hey all,

Is it possible to facilitate compression on a main CSS file, with nesting or another style on a secondary CSS file?

Whether it for production or live amends. I often have to provide compressed and nested typed of the same & secondary CSS Files. At present I am changing the config.rb if and when...


scottdavis:
Needs to get rebased against master so we can review the changes that were made. I'm closing this PR when you get it rebased then resubmit a new PR please.

PR #1334, opened 2013-07-15
Fix woff mimetype to be consistent with latest W3 spec.

abierbaum:
The W3 spec now says that mimetype for woff fonts should be: application/font-woff

http://www.w3.org/TR/WOFF/#appendix-b

See:
- http://zduck.com/2013/google-chrome-and-woff-font-mime-type-warnings/
- http://stackoverflow.com/questions/16704967/resource-interpreted-as-font-but-transferred-with-mime-type-application-x-font-w

note: This was changed before in compass about a year and a half ago: https://github.com/chriseppstein/compass/pull/755  

Now it seems to need changed again.

### DIFF
--- a/lib/compass/sass_extensions/functions/inline_image.rb
+++ b/lib/compass/sass_extensions/functions/inline_image.rb
@@ -44,7 +44,7 @@ private
     when /\.ttf$/i
       'font/truetype'
     when /\.woff$/i
-      'application/x-font-woff'
+      'application/font-woff'
     when /\.off$/i
       'font/openfont'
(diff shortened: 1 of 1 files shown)


chriseppstein:
I'm concerned about what this means for legacy browsers. do they still need `woff` or `x-font-woff`?

chriseppstein:
according to the referenced article, this change will cause warnings in some older browsers, but should keep working if we change it, so I'll accept the patch.

chriseppstein:
I landed this onto master.

Issue #1335, opened 2013-07-16
Update html5 element list to contain <main> tag


daphotron:
The html5shiv has been updated to contain support for the main tag, and http://html5doctor.com/the-main-element/ has been written about it. Is it time to add it to compass?

scottdavis:
^^^

PR #1336, opened 2013-07-16
Add <main> tag as a new html5 block level element

daphotron:
See Issue #1335

### DIFF
--- a/lib/compass/sass_extensions/functions/display.rb
+++ b/lib/compass/sass_extensions/functions/display.rb
@@ -2,7 +2,7 @@ module Compass::SassExtensions::Functions::Display
   DEFAULT_DISPLAY = {
     :block => %w{address article aside blockquote center dir div dd details dl dt fieldset
                   figcaption figure form footer frameset h1 h2 h3 h4 h5 h6 hr header hgroup
-                  isindex menu nav noframes noscript ol p pre section summary ul},
+                  isindex main menu nav noframes noscript ol p pre section summary ul},
     :inline => %w{a abbr acronym audio b basefont bdo big br canvas cite code command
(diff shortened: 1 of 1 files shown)


mirisuzanne:
Thanks for the patch! But currently this is breaking the tests. Do you mind updating the tests and re-submitting?

daphotron:
@ericam Oh definitely! :) So I received this error `compass/test/helpers/test_case.rb:42:in `setup': wrong number of arguments (2 for 0) (ArgumentError)` after running `bundle exec rake test features` but I found that by commenting out the setup block as mentioned in this closed pull request https://github.com/rubiii/compass/commit/e5d9fdff8e04916856889280609d82173d0d7d75 allowed the bundle test to run. 

After getting that hurdle over with...I wasn't sure how to recompile the css files inside the test folder. I tried running `compass compile --force` and other such commands inside individual folders inside `compass/test/fixtures/stylesheets` but noticed that they were 1) going to a tmp directory and 2) not picking up the latest changes I made to the `html5-block` variable. Forgive my noobness but I couldn't find any good docs about editing tests other than http://compass-style.org/help/tutorials/contributing/#running-tests and the docs made it seem that by running the test, the test would be "changing the expected css output to make the test pass". Is there a command I'm missing or more documentation on how to update the stylesheet tests?

mirisuzanne:
I'm not sure if there are ruby unit tests that need to be updated, but for the css output tests, you have to manually change the expected output. When you `rake test` - the tests are compiled into the `saved` folder and then compared to the expected output in the `css` folder. You need to make sure the saved output is correct, and then update the expected css to match. Then run the tests again to make sure they pass.

chriseppstein:
This was implemented in 32b3a49f55b06691dcdd9f135ba2a95184a17ca9.

Issue #958, opened 2012-07-10
text-shadow produce broken syntax


louim:
Since The update to version 0.12.2, text-shadow and single-text-shadow produce invalid syntax:

this:

@include text-shadow(1px,2px,2px,rgba(#000,0.25));

produce :

text-shadow: 1px 2px 2px #AAA, 1px 2px 2px rgba(0, 0, 0, 0.25) #AAA;
text-shadow: 1px 2px 2px rgba(0, 0, 0, 0.25) #AAA, 1px 2px 2px rgba(0, 0, 0, 0.25) #AAA;

Instead of the normal syntax.

StanAngeloff:
Use spaces instead of a comma, does that help? I also had something similar when I was including Bootstrap after Compass, believe at one point it had it's own `text-shadow` @-mixin.

Also note `single-text-shadow` added `$spread`, you'd want to include it before your colour or use: `$color:` in front, e.g.,

``` scss
@include single-text-shadow(1px, 2px, 2px, $color: rgba(#000,0.25));
```

louim:
Using spaces only produce only one line of text-shadow, but the declaration is still invalid :

``` css
text-shadow: 1px 2px 2px rgba(0, 0, 0, 0.25) 1px #AAA;
```

using the $color fixed the problem.
Maybe this should be explained in the documentation?

silvenon:
The documentation has an [example](http://compass-style.org/examples/compass/css3/text_shadow/) on how to use the mixins. Try putting color as the first item in the list, does it still produce invalid CSS?

patrikmakrai:
I also experienced this bug, this is my code:

```
@include single-text-shadow(-1px, -1px, 0px, 1px, $color: #414141);
```

And this is the output:

```
text-shadow: -1px -1px 0 #414141, -1px -1px 0 1px #414141;
text-shadow: -1px -1px 0 1px #414141, -1px -1px 0 1px #414141;
```

I realize it is duplicated due to the "spread" fix, but the syntax is still invalid. I even tried it with the "$color: " declaration, but the problem still persists.

kopijunkie:
Was experiencing this bug and have found that writing it this way gives the right output:

``` scss
@include single-text-shadow(1px, 1px, 2px, $color: #fefefe);
```

PR #1338, opened 2013-07-16
Fix broken `text-shadow` syntax

joeyhoer:
A small bug, the `@` for the `else` statement was omitted, and causing the else statement to always be evaluated.

Proposed fix for #958

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_text-shadow.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_text-shadow.scss
@@ -41,7 +41,7 @@ $default-text-shadow-spread:   false  !default;
       $has-spread: true;
       $shadows-without-spread: append($shadows-without-spread, nth($shadow,1) nth($shadow,2) nth($shadow,3) nth($shadow,5));
       $shadows: append($shadows, $shadow);
-    } else {
+    } @else {
       $shadows-without-spread: append($shadows-without-spread, $shadow);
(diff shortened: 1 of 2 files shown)


mirisuzanne:
Thanks! This looks like a great fix, but we'll need the tests updated as well. Can you take a look at that and update the pull request?

joeyhoer:
AFIK, the current tests, when used with this commit, will produce the expected/valid CSS.

However, the current tests, when run through the current code will produce unexpected/invalid syntax due to the bug.

I've run the tests and everything appears to be in working order.

mirisuzanne:
The way compass tests work, you need to update what is expected under `test/fixtures/stylesheets/compass/css/text_shadow.css`. Currently it shows:

``` css
.multiple-text-shadows-with-spread {
  text-shadow: 4px 4px 10px #444444, 4px 4px 10px 1px #444444, 2px 2px 5px #222222, 2px 2px 5px 3px #222222;
  text-shadow: 4px 4px 10px 1px #444444, 4px 4px 10px 1px #444444, 2px 2px 5px 3px #222222, 2px 2px 5px 3px #222222; }
```

That is the incorrect output - due to the bug. Clearly someone wasn't paying attention when they wrote that test. Update that to match the correct output, and everything should work.

chriseppstein:
Thanks. This file got basically re-written for 1.0 and this bug is no longer present.

PR #1346, opened 2013-07-22
Fix #949.

### LINKED ISSUE #949: Compass does not compile with Unix line-endings on Windows
Compass version: 0.12.2
Ruby version: tested on 1.9.3 and 1.8.7
Windows version: Windows XP and Windows 7

Steps to recreate:
1. config.rb file in a compass project sets sass_options = {:unix_newlines => true}
2. on the command line in Windows, watch the project using compass watch and have the files recompile

What I expect to happen: the files get compiled with unix line endings.

What actually happens: the files compile with Windows line endings.

We can see the difference because the members on our team who are on a Mac compile and commit to the same SVN repo without issue, but when a Windows user watches the same set of files, they get re-compiled with no differences other than line endings.

johnmave126:
Fix #949.
Use binary mode to write file if unix_newlines is set.
I don't know if there will be any side effect if using binary mode to write file. At least I tested it on my machine and no side effect up to now.

### DIFF
--- a/lib/compass/compiler.rb
+++ b/lib/compass/compiler.rb
@@ -141,7 +141,7 @@ module Compass
         end
       end
       duration = options[:time] ? "(#{(css_content.__duration * 1000).round / 1000.0}s)" : ""
-      write_file(css_filename, css_content, options.merge(:force => true, :extra => duration))
+      write_file(css_filename, css_content, options.merge(:force => true, :extra => duration), sass_options[:unix_newlines])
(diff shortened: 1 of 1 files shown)


johnmave126:
oops...seems unit test failed...
I am going to look into it a little bit.

johnmave126:
It seems this failure is not my fault?
Fail on stale HEAD also?....

scottdavis:
stable has been failing for ahile due ot a botched PR. Please make changes against master

PR #1352, opened 2013-08-04
New mixin use-baseline in vertical rhythm module

gabrielnau:
Discussed with @ericam here : https://groups.google.com/forum/#!topic/compass-users/NecGNw31Rzc

The mixin takes a variable argument list of base font size, base line height and changes the context for the given block.

@debug-grid-background needed an additional argument, as it relies on the current value of $base-font-size which may be the wrong one. I didn't find any other mixin of function to patch.

I also made $base-leader and $base-half-leader function since they can mistakenly be used in the alternative baseline context without taking it into account. If this change is a problem, we should may be warn the user in the doc ?

Concerning the tests, I just added the new mixin in test/fixtures/stylesheets/compass/sass/vertical-rhythm, is that enough ?

### DIFF
--- a/frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
+++ b/frameworks/compass/stylesheets/compass/typography/_vertical_rhythm.scss
@@ -131,12 +131,12 @@ $relative-font-sizing: if($rhythm-unit == px, false, true);
 // for rem output.
 // Include the $img argument if you would rather use your own image than the
 // Compass default gradient image.
-@mixin debug-vertical-alignment($img: false) {
+@mixin debug-vertical-alignment($img: false, $font-size: $base-font-size) {
   @if $img {
     background: image-url($img);
   }
   @else {
-    @include baseline-grid-background(rhythm(1, $base-font-size));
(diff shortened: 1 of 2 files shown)


mirisuzanne:
This will need updated output css for the tests to work.

chriseppstein:
Ping.

chriseppstein:
I'd like to wait on this until after 1.0

gabrielnau:
I really apologize, I completely forgot about this pull request and I know this can be truly annoying for maintainers to have these pending subjects.

If I understand, you don't want to merge this (with missing tests) for 1.0, that's right ?

Undistraction:
Is this still being considered or has it been abandoned?

gabrielnau:
Since this PR, I more or less leaved this field. I have no idea how much this PR diverged from master and so I cannot tell how much work is required to finish it or not. As @chriseppstein said, it cannot be merged either way before 1.0 and this version is still in alpha (1.0.0.alpha.21).

Undistraction:
@gabrielnau Thanks for the update. I actually thought Compass had reached 1.0 a while ago.

chriseppstein:

PR #1359, opened 2013-08-13
fixed a bug with inline-image in opera and IE

MindFreeze:
When you specify the mimetype with quotes like so: inline-image('content-bgr.png', 'image/png'); it is written in the css with double quotes and this breaks it in Opera(found in 12) and IE (found in 8 and 9). This fixes it.


chriseppstein:
I landed this on master.

PR #1362, opened 2013-08-14
Remove -webkit-flexbox

mgol:
There's no such thing as `-webkit-flexbox`. This pull request disables `-webkit`-prefixing the 2012 Working Draft flex properties.

Also, browser support information is updated - I added info about IE & Safari and corrected Firefox info (the CR Flexbox has been available in Firefox since version 22, not 20).


mgol:
Test failures are not mine, it all fails on current `master` as well.

chriseppstein:
Thanks for the patch, The flexbox module was completely rewritten for 1.0 and no longer contains -webkit-flexbox.

PR #1372, opened 2013-09-06
Make it work with latest sass 3.3.0 alpha

zemd:


scottdavis:
This is causing CI failures.

zemd:
There is incompatible code with sass 3.3.0 in compass\lib\compass\sass_extensions\sprites\image_methods.rb because in version 3.3 base class for sass script values are not inherited from Sass::Script::Tree::Node

```
def image_for(name)
          if name.is_a?(Sass::Script::String)
            name = name.value
          end
          name = name.to_s
          @images.detect { |img| img.name == name}
         end
```

It returns nill and sprites tests fail https://travis-ci.org/hunterman/compass/builds/11089603 
My last commit fixed issue with 3.2 compatibility and I think sprite fixes should be in independent pull request.

zemd:
It seems that these are sass.alpha problems, tried to explain that I found here https://github.com/nex3/sass/issues/913
sorry for my poor english

chriseppstein:
I sorted all this out. Thanks for the patch, tho.

PR #1375, opened 2013-09-10
Fix subbing similar path names while generating sprite map

puffy:
There is a problem when compass project have multiple directories with assets and this directories have similar names, e.g.:

```
/long/path/project/app/assets/blocks
/long/path/project/app/assets/blocks-mobile
```

Without this fix, for the second directory name `sprite.include?(path)` return `true` when `path` is `project/app/assets/blocks` and sprite generation will be fail.


scottdavis:
Needs a test case

puffy:
Wrote a test

PR #1377, opened 2013-09-13
Added default box sizing option.

martinbavio:


martinbavio:
This is my first attempt to do a PR into Compass, if I've done something wrong, please let me know!

mirisuzanne:
Thanks! Looks like you got tests in place and everything, which is great.

My only concern is: why not use the built-in sass method for defining defaults?

``` scss
@mixin box-sizing($box-model: $default-box-sizing) { // default set in the argument list
  $box-model: unquote($box-model);
  @include experimental(box-sizing, $box-model, $box-sizing-support...);
 }
```

martinbavio:
Yeah, that's definitely better.

I saw other Compass files and they were not using the default param option, like:
1. Using set-arglist-default:
   https://github.com/chriseppstein/compass/blob/master/frameworks/compass/stylesheets/compass/css3/_background-size.scss
2. If and Else:
   https://github.com/chriseppstein/compass/blob/master/frameworks/compass/stylesheets/compass/css3/_background-clip.scss

What do you suggest?

mirisuzanne:
Yeah, Sass doesn't have a good way of handling defaults for an arglist - where the argument can take any number of values. Looks like we've used a few different techniques to get around that, and we should get our act together. :)

In the case of box-sizing, we don't need to handle an arbitrary arglist, so the standard Sass method should work fine.

martinbavio:
Mixin updated. :)

martinbavio:
@ericam What's the status of this? Arent you merging any PR until certain date or something?

mirisuzanne:
I didn't get that memo. :)

Looks like there's an issue with the tests — i think you have the input backwards from the output.

I also think the default should probably be `border-box` — since `content-box` is already the default on all elements, and most people are wanting that changed to `border-box`.

chriseppstein:
I'm torn on whether our defaults should be the browser default or "the most common thing people want to set it to". I feel like the browser default is best for the principle of least surprise for new users, but this is not how we've been doing it and consistency is the principle of least surprise for existing users.

So I think we should be consistent, make the default `border-box` and maybe down the road we'll deprecate our defaults and change to the browser default.

martinbavio:
I’ve updated default value as @chriseppstein has sugested. Also, I’ve edited the tests which should be fine now, what do you think?

chriseppstein:
So many changes to the box sizing module. I just added this myself: 2cd344da24977a9973ea69dcdc8168fee709bfa4

martinbavio:
And there goes my chance of having a merged pull request on the open source project I like the most on the web. :(

mirisuzanne:
@mbavio Don't give up yet! The new approach to things in Compass 1.0 should make patches easier than ever - and the new release cycle should get things pulled in more quickly. I'm sure we'll continue making mistakes and missing obvious features, so your contributions will be welcome. :)

PR #1380, opened 2013-09-17
Allow compass compile to traverse symlinks. Fixes #1379

### LINKED ISSUE #1379: Compass compile does not traverse symlinked directories
Sometimes asset pipeline setups require directories to be symlinked from a working directory to a public directory, for example when using mixins defined in bower'd components.

Compass currently doesn't support this because `Dir.glob('**')` does not traverse symlinks to prevent infinite loops. 

Sprokets had a similar issue and solved it by allowing symlink traversal for up to one level https://github.com/sstephenson/sprockets/pull/189.

xzyfer:
Sometimes asset pipeline setups require directories to be symlinked from a working directory to a public directory, for example when using mixins defined in bower'd components.

Compass currently doesn't support this because `Dir.glob('**')` does not traverse symlinks to prevent infinite loops. 

Sprokets had a similar issue and solved it by allowing symlink traversal for up to one level https://github.com/sstephenson/sprockets/pull/189. I've ported their patch to compass.

I'm not a ruby developer by trade so I image this could be made more terse.


chriseppstein:
My guess is the sass compiler has a similar issue. I'd really like to standardize on a single implementation for both the sass compiler is quite good now and can replace the compass compiler, IMO.

I'd like to wait to merge this until after 1.0.

chriseppstein:
Compass uses the Sass compiler now. If this bug exists there, we should fix it in Sass.

PR #1392, opened 2013-09-30
added always even positions to the smart-layout with an extra padding of 2

afewyards:
Added the always even positions to the smart layout to fix the ghosting of retina sprites. they get divided and if there positions is odd the position will be not on a int. This will fix that


scottdavis:
I like the thought behind this patch, but it needs the test cases updated before i accept it.

afewyards:
ahh, ok will do that when I got more time.. prob somewhere later this week.

chriseppstein:
ping?

afewyards:
ok, took a while but I updated the test cases and altered the sprite generation a bit to have a better/good result.

afewyards:
I see that "rcov" fails in the 1.9+ builds.. as that needs to be converted into "simplecov" that is not part of this pull request.

chriseppstein:

@scottdavis has refactored sprites considerably, we need to rebase this work to take his recent changes (soon to land on master) into account.

chriseppstein:
Should be on top of the changes coming in #1771.

PR #1409, opened 2013-10-14
Auto-requiring compass command-line extensions & loading ad-hoc command-line extensions

JohnRiv:
As Chris Eppstein & I discussed at SassConf, this is a much simpler, more efficient, and more powerful solution than my [original proposal](https://github.com/chriseppstein/compass/pull/1261) to fix issue #1053 :-)

Assuming Compass command line extension are created with the naming convention of always being prefixed with "compass-", this will allow you to use that Compass command line extension by just installing the gem. No need for a command_extensions property in the config.rb, nor an extra "require" statement.

This also works with ad-hoc extensions as well (that's what line 17 is for)

If this is approved, then [compass-csslint](https://github.com/Comcast/compass-csslint) & [compass-csscss](https://github.com/Comcast/compass-csscss) will be much easier to use, and it will hopefully open the door up for others to create Compass command line extensions.

### DIFF
--- a/lib/compass/commands.rb
+++ b/lib/compass/commands.rb
@@ -10,3 +10,9 @@ require 'compass/commands/registry'
 ).each do |lib|
   require "compass/commands/#{lib}"
 end
+
+Gem.find_files("compass-*").map{|f| File.basename(f, ".rb")}.each do |compass_extension|
+  require compass_extension
+end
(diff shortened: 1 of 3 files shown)


chriseppstein:
Closing this PR, it's been folded into #1439 and will land soon.

Issue #1422, opened 2013-10-24
Any interest in better Sprite bin packing?


kiafaldorius:
I'm thinking I might want to spend an afternoon next week looking for and/or implementing a better packing algorithm for the sprites.

Is there any interest in it? Do you think it would be worth while to have a much slower, but much less-whitespacey spriter?

Might still spend a couple hours on it regardless of your answer...just wanted to hear your take on it, seems the spriting stuff hasn't been touched in 2 years...

scottdavis:
have you used smart layout? http://compass-style.org/help/tutorials/spriting/sprite-layouts/#smart

kiafaldorius:
I saw it last night when I skimmed through the code. But...I don't think it's very smart.

Here's a quick example, each colored square is a file:

![mysprite-s792cf60370](https://f.cloud.github.com/assets/119071/1404105/b8dac158-3cfe-11e3-9572-3c2a7c6be97f.png)

Should be able to do better than that. Maybe not, I'm not sure. But I'm willing to give it a shot. In the interest of not re-inventing the wheel, I'd probably go searching up other css spriters first, someone's got to have figured something out by now...

scottdavis:
so we have this idea of efficiency which will make the packer scan more and do what you want I think. this could be pulled out into a config and or refactored. I'm not against making this better but, just realize there is a trade off on size / speed.

https://github.com/chriseppstein/compass/blob/stable/lib/compass/sass_extensions/sprites/row_fitter.rb#L36-L38

PR #1433, opened 2013-10-31
New layout for sprites, heavily "packed"

kiafaldorius:
much more compact than "smart" when images' height and width are differently sized.

All credit goes to Jakes Gordon for the algorithm (and most of the code).

PackFitter algorithm description:
http://codeincomplete.com/posts/2011/5/7/bin_packing/

code adapted from:
https://github.com/jakesgordon/sprite-factory/blob/master/lib/sprite_factory/layout/packed.rb

See Issue #1422
## Comparison images

layout: smart
mysprite-s792cf60370

layout: packed
mysprite-sf32dae54b5


kiafaldorius:
Haven't done benchmarks on it yet, but for the sample rectangles I gave, the generation times are pretty much identical. with the packed being 20 bytes smaller (282 vs 262).

I expect real world cases will yield much better compression since it really does pack quite a bit better in exchange for more memory usage (could be definitely be optimized further, by dropping the generated hash tree after packing completes).

scottdavis:
Looks good I'm ok with just replacing smart packing all together ping @chriseppstein

arekbartnik:
Did you consider spacing for sprite's elements?

kiafaldorius:
Spacing/margins wasn't built in because I didn't see any functionality for that in smart packing. It just pulls Image#height and Image#width to do the calculations.

It's simple enough to add though, just need to add the margin pixels to the height and width; the core calculations will still remain the same.

er1z:
@kiafaldorius - the image weight isn't everything. The less pixels on the image are, the less memory they take (I mean browser right now).

Any picture - nevertheless compression - is decoded to the bitmap in memory. Small sprites - no problem. But the larger ones...

chriseppstein:
This looks great.

How does this algorithm's performance compare to the current smart packing algorithm with dozens of images in the sprite?

Assuming comparable performance, I'm ok with simply bumping [the sprite version](https://github.com/chriseppstein/compass/blob/master/cli/lib/compass/sass_extensions/sprites/sprite_methods.rb#L8) and making this the smart algorithm.

The fact that spacing wasn't supported in the smart algorithm was a bug. I'd like to not carry that bug into this algorithm.

scottdavis:
Actually the spacing thing wasn't a bug it was just not implemented. there were feature requests but no one ever picked up the ball and finished it. The docs made it clear that configurations for smart pack spacing, padding etc were not supported :)

kiafaldorius:
When I find some time--possibly this weekend--I'll hook up spacing and throw out a few benchmarks with random icons from the openicon library. I'm confident the height and width will be smaller, whether the actual filesize is substantially smaller we'll have to see.

Since this thing benefits from proper sizing, I'm considering allowing per-side spacing, eg: spacing-top, spacing-bottom, spacing-left, spacing-right.

kiafaldorius:
Smart Sprite:
46055
real    0m12.532s
user    0m3.170s
sys 0m3.907s
mysprite-s5743b235ff

Packed Sprite:
45937
real    0m11.299s
user    0m2.780s
sys 0m4.467s
mysprite-sd573c13ba3

Smart Sprite:
128590
real    0m12.422s
user    0m4.857s
sys 0m3.460s
mysprite-sbb0f6f07d7

Packed Sprite:
129826
real    0m12.349s
user    0m4.463s
sys 0m3.727s
mysprite-s68aa5a03ee

Smart Sprite:
240453
real    0m14.989s
user    0m5.997s
sys 0m4.297s
mysprite-s53debc97e6

Packed Sprite:
244155
real    0m14.353s
user    0m5.920s
sys 0m3.980s
mysprite-s96c779cbab

So it appears extended square packing might not play so well with png's compression. Which I somewhat expected...But packed might be slightly faster, which is not what I would have thought.

kiafaldorius:
So, with that in mind, it may be wise to keep them separate.

Pack's pixel counts are always less than the smart, so presumably it would still be much more memory efficient to render...at the cost of around 1-2% more filesize (filesize -> bandwidth, so slightly more transit time) for large file counts compared to smart. For two dozen or so, packed is still usually smaller than smart.

I've just pushed my current changes. Commits 68a1f2b and 3ffd7b9 add spacing to the existing smart layout without touching the packed version, so should be safe to cherry pick if you decide not to take the new algorithm. (I had lots of test failures/errors show up when trying to bump the sprite version, probably to do with hardcoded checksums. I'll leave that part to someone in the dev team.)

chriseppstein:
I like this, but we need to get it rebased on top of 1.0.

chriseppstein:
Actually, on top of #1771 after it lands.

kiafaldorius:
Hey, sorry for late reply, been busy lately.

What's the status of #1771 ? Can I start rebasing it or will I need to wait until it gets into production? (It appears that pull is not passing travis?)

scottdavis:
@kiafaldorius you can rebase aginst that the tests are erroring atm due to some travis weirdness but they pass localy just fine

kiafaldorius:
Merged in the changes from decomposed_sprites branch in #1771 The sprite tests pass. I had to update the Gemfile's ffi though, 1.9.4 got yanked from rubygems, replaced with 1.9.5 @scottdavis

scottdavis:
Awesome

Issue #1225, opened 2013-03-23
$experimental-support-for-svg with radial-gradient generates <radialGradient> svg element with fixed radius - 100px


okendoken:
When using compass radial-gradient mixin with $experimental-support-for-svg enabled it produces incorrect svg image where radial gradient radius is set to 100 instead of 100%.

``` xml
<?xml version="1.0" encoding="utf-8"?> 
<svg version="1.1" xmlns="http://www.w3.org/2000/svg">
  <defs>
    <radialGradient id="grad" gradientUnits="userSpaceOnUse" cx="10%" cy="0" r="100">
      <stop offset="0%" stop-color="rgba(137, 111, 82, 0.94)"/>
      <stop offset="20%" stop-color="rgba(194, 123, 108, 0.94)"/>
      <stop offset="100%" stop-color="rgba(238, 231, 189, 0.94)"/>
    </radialGradient>
  </defs>
  <rect x="0" y="0" width="100%" height="100%" fill="url(#grad)" />
</svg> 
```

The fix is to change 100 to 100%.
I am going to learn how to contribute to your repo and probably fix it by myself.

PR #1436, opened 2013-11-04
Use percentage for radius of radial SVG gradient.

koenpunt:
As suggested in #1225


chriseppstein:
This breaks tests, and I'm not sure whether it's a backwards compatible change.

chriseppstein:
I merged this to master. Hopefully no regressions will result :four_leaf_clover: 

http://git.io/28HHzg

koenpunt:
I noticed about the tests, but for some reason I couldn't get the fixtures updated with the new data as there were constant mismatches..

PR #1440, opened 2013-11-06
debug-vertical-alignment() should respect $round-to-nearest-half-line

acdlite:
If `$round-to-nearest-half-line` is true, the grid lines generated by `debug-vertical-alignment()` should be positioned at every half of a rhythm unit, rather than a full one.


chriseppstein:
merged as 05f7c2a95cb7724ae5198fd3ffc1b949733a5288

PR #1442, opened 2013-11-07
fix copy-paste errors in sprite tutorial

mprins:
seems as if one example containing 'icon' was copy-pasted all over the place but the example code wasn't


scottdavis:
nice catch @chriseppstein :+1:

chriseppstein:
merged. Thanks!

PR #1444, opened 2013-11-12
improved support for inline-font-files(), adds font-formats() method

eoneill:
this modifies `inline-font-files` so it can take the same syntax as `font-files`

``` scss
// old syntax (still works)
$files: inline-font-files('path/to/file.ttf', truetype);

// new syntax (without the format specified)
$files: inline-font-files('path/to/file.ttf');
```

this also fixes an issue where `font-files` couldn't take an unquoted string (because of the regex pattern used):

``` scss
$files: font-files( unquote('path/to/file.ttf') );
```

and added a `font-formats` method that will return a list of formats given the above syntaxes:

``` scss
$formats: font-formats('path/to/file.ttf'); // truetype
$formats: font-formats('path/to/file.ext', 'otf'); // opentype
$formats: font-formats('path/to/file.ttf', 'path/to/file.ext', 'otf'); // (truetype, opentype)
```

each of these methods (`font-files`, `inline-font-files`, `font-formats`) each uses shared logic for deriving the format `with_each_font_file`

### DIFF
--- a/lib/compass/sass_extensions/functions/font_files.rb
+++ b/lib/compass/sass_extensions/functions/font_files.rb
@@ -9,33 +9,55 @@ module Compass::SassExtensions::Functions::FontFiles
     :eot => 'embedded-opentype'
   }
 
+  def font_formats(*args)
+    formats = []
+    with_each_font_file(*args) do |path, type|
+      formats << Sass::Script::String.new(type)
+    end
+    return Sass::Script::List.new(formats)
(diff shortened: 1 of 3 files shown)


chriseppstein:
merged as de25ffbb951db662fe01f81b50f405b3d1e48400 and 36c32931d0ee5a858f8ea76f579718cba6fd4d53

Issue #1005, opened 2012-08-28
Enabling compass sprite classes to be output as placeholders


superbland:
Currently this code:

```
@import "site/*.png";
@include all-site-sprites;
```

Produces an output that contains a list of classes relating to the background image and positions that can be extended. Since the recent addition of placeholders and the move to prevent extending from within media directives, it will be impossible to extend these classes without potentially having massive amounts of duplicate code within the media query.

Using placeholders instead of classes (or allowing a choice) would mean that a mixin containing sprite css could be included without unnecessary duplication.

silvenon:
Unfortunately, Sass doesn't seem to be able to extend placeholders beyond the `@media` block (and probably classes too, I haven't tried it yet). Nevertheless, it's a great suggestion, I'll implement it when I find the time.

Yeah, there will be a choice, the default will be `class` for backwards compatibility.

aaronjensen:
+1

silvenon:
I forgot about this... Don't count on me because I don't use Compass anymore.

scottdavis:
Well that is sad news

silvenon:
Probably not permanently, it's just because it's so messy in its current state. I'll see if I'll go on a fixing spree later.

morewry:
In the meantime, a workaround option for using placeholder classes with Compass sprites https://gist.github.com/morewry/4151748

scottdavis:
So are you advocating that the default be placeholders? this is a pretty simple change just need to get the option of those pushing for it.

morewry:
Placeholders as the default? Maybe; don't feel strongly about which the default should be. Opting for placeholders instead of classes? Yes, I'd like to see that option.

scottdavis:
Anyone feel like doing a pull request?

aaronjensen:
Is there a decision on how it should be done? I like the idea of placeholders by default but that would be a breaking change for many. It probably makes more sense to do classes by default for the `@include all-site-sprites;` but placeholders by default for `@import "site/*.png";` since maybe people aren't using the initial class that the import creates other than via mixin?

scottdavis:
Its ok to do a breaking change with a new version release (0.13). So I'm ok setting placeholders to default. The framework functions and mixins would need to be updated along with the content.erb in the sprite importer folder, and all the test files.

lucasconstantino:
Any news on this?

morewry:
[...]

PR #1448, opened 2013-11-17
Use placeholders for sprite selectors

lucasconstantino:
This pull request is an answer to the issue #1005.

It's important to note that this pull request should not modify any current default behavior for sprites, meaning every code being used today will continue to work perfectly fine.

What this does is give the developer the option to create the sprite selectors as placeholders, so that they won't be emitted to the CSS unless the developer makes use of the sprites by extending these selectors.

There are two ways to making use of this functionality:
1. **Set the newly introduced variable `$all-sprites-as-placeholders` to 'true'**: this will cause all sprite insertions that follow this definition to be output as placeholder;
2. **Set the dynamic sprite specific variable `$<sprite-name>-placeholder-sprite` to 'true'**: this will cause only the specific sprite to be output as placeholder.

Note that the second method overwrites the first, so that you can use them in combination say, for example, to output all sprite as placeholders but one specific, by doing the first option and doing the reverse of the second option.

### DIFF
--- a/frameworks/compass/stylesheets/compass/utilities/sprites/_base.scss
+++ b/frameworks/compass/stylesheets/compass/utilities/sprites/_base.scss
@@ -68,12 +68,15 @@ $default-sprite-separator: "-" !default;
 // Determines if you want to use placeholder instead of class selectors.
 $all-sprites-as-placeholders: false !default;
 
-// Generates a class for each space separated name in `$sprite-names`.
-// The selector will be of the form .<map-name>-<sprite-name>.
+// Generates a selector for each space separated name in `$sprite-names`.
+// The selector will be of the form <base-selector>-<map-name>-<sprite-name>,
+// where <base-selector> defaults to `.`.
 //
 // If a base class is provided, then each class will extend it.
 //
 // If `$dimensions` is `true`, the sprite dimensions will specified.
+// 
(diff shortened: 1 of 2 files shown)


Frikki:
Good idea.

revagomes:
+1

chriseppstein:
Instead of adding a whole new option, I think it would be better to respect the `$<folder>-prefix` variable and have it passed to the `$prefix` argument of the mixins. And then to have the handling of that variable change to allow the inclusion of either a class, placeholder, or id prefix; when none is provided just default to a class selector.

I think this allows people more flexibility in the name of the prefix, keeps the API stable and allows for ID selectors too.

Example:

``` scss
$icon-2x-prefix: '%retina-icon';
@import "icon-2x/*.png";
@include all-icon-2x-sprites;

// $icon-2x-sprite-base-class == "%retina-icon-sprite"
// creates %retina-icon-save, %retina-icon-undo, etc."
```

``` scss
$icon-2x-prefix: '#retina-icon';
@import "icon-2x/*.png";
@include all-icon-2x-sprites;

// Because an ID base class doesn't make sense, default to placeholder for that situation.
// $icon-2x-sprite-base-class == "%retina-icon-sprite"
// creates #retina-icon-save, #retina-icon-undo, etc."
```

Issue #1454, opened 2013-11-26
No special treatment for Webkit in Transition mixin


JoernBerkefeld:
From my experience, current webkit browsers all support adding the delay directly into the -webkit-transition property. The current chrome even does no longer understand the -webkit-transition-delay property and flags it as invalid

http://beta.compass-style.org/reference/compass/css3/transition/#mixin-transition

Can be fixed by removing all special webkit lines.

Issue #1446, opened 2013-11-13
Compass > 0.13.alpha.4 screws up Transitions


pyronaur:
### Sass:

```
.js__trans
    +transition(transform 400ms cubic-bezier(.01,.99,.62,.91), opacity 300ms cubic-bezier(.01,.99,.62,.91))
```
### Compass 0.13.alpha.4: ( as I expect it to be )

```
.js__trans {
  -webkit-transition: -webkit-transform 400ms cubic-bezier(0.01, 0.99, 0.62, 0.91), opacity 300ms cubic-bezier(0.01, 0.99, 0.62, 0.91);
  -moz-transition: -moz-transform 400ms cubic-bezier(0.01, 0.99, 0.62, 0.91), opacity 300ms cubic-bezier(0.01, 0.99, 0.62, 0.91);
  -o-transition: -o-transform 400ms cubic-bezier(0.01, 0.99, 0.62, 0.91), opacity 300ms cubic-bezier(0.01, 0.99, 0.62, 0.91);
  transition: transform 400ms cubic-bezier(0.01, 0.99, 0.62, 0.91), opacity 300ms cubic-bezier(0.01, 0.99, 0.62, 0.91); }

```
### Compass 0.13.alpha.10

```
.js__trans {
  -moz-transition: -moz-transform 400ms cubic-bezier(0.01, 0.99, 0.62, 0.91) opacity 300ms cubic-bezier(0.01, 0.99, 0.62, 0.91);
  -o-transition: -o-transform 400ms cubic-bezier(0.01, 0.99, 0.62, 0.91) opacity 300ms cubic-bezier(0.01, 0.99, 0.62, 0.91);
  -webkit-transition: -webkit-transform 400ms cubic-bezier(0.01, 0.99, 0.62, 0.91) opacity 300ms cubic-bezier(0.01, 0.99, 0.62, 0.91);
  transition: transform 400ms cubic-bezier(0.01, 0.99, 0.62, 0.91) opacity 300ms cubic-bezier(0.01, 0.99, 0.62, 0.91); }
```

Anahkiasen:
...I don't see any difference, what's wrong with the current one ?
EDIT: Oh wait, the missing comma between the transitions, got it.

eikaramba:
Yes i can confirm this(was actually searching if someone else has the problem). Multiple transitions cannot be seperated by commas, or more accurate, compass removes the commas.

scottdavis:
[...]

PR #1455, opened 2013-11-26
fixed #1450, #1454, #1446

### LINKED ISSUE #1450: multiple transitions are missing "," in between
seen in 0.13.alpha.10:

@include transition(opacity 0.5s ease, visibility 0.5s);

should output:
transition: opacity 0.5s ease, visibility 0.5s;
instead it outputs:
transition: opacity 0.5s ease visibility 0.5s;

--> without the comma!

I omited the prefixed versions here but the problem is the same for those

JoernBerkefeld:
changed transition lists to comma-lists and removed all webkit-special code because the current Chrome does not understand it


cimmanon:
Does this still work in all relevant Webkit browsers still in use (Safari, Blackberry, Android)?

JoernBerkefeld:
Failed to test that. Will add test results

JoernBerkefeld:
hmm, that test seems to take more time than I thought. will reduce the update to a fix 1446 and 1450

PR #1457, opened 2013-11-27
really fixed #1450, #1454, #1446

### LINKED ISSUE #1450: multiple transitions are missing "," in between

JoernBerkefeld:
fixes multiple transitions that were missing commas in between

### DIFF
--- a/frameworks/compass/stylesheets/compass/css3/_transition.scss
+++ b/frameworks/compass/stylesheets/compass/css3/_transition.scss
@@ -118,38 +118,49 @@ $transitionable-prefixed-values: transform, transform-origin !default;
 }
 
 @mixin transition($transitions...) {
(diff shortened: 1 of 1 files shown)


chriseppstein:
I squashed this into a single commit, updated the tests and fixed a bug that this introduced :) Thanks for the patch. See fc360d24d09223455a08bbafe360d62c125a36d3 & 74e2bcf902f344c9049df4189dbf870336ae32e9.

PR #1465, opened 2013-12-04
enabled -ms prefix in user-select mixin

whazap:
just a minor change to enable user-select for IE10+
http://ie.microsoft.com/testdrive/HTML5/msUserSelect/


chriseppstein:
This is already working on master. Please try the latest preview release.

whazap:
whups! of course. thanks

PR #1469, opened 2013-12-06
zero length should not have units assigned

mprins:
lengths of 0 do not require any unit assigment.


chriseppstein:
This will break the tests. Please update them. Thanks!

mprins:
@chriseppstein I updatet the tests.

As a side note: I was wondering if  "5. Running behavior tests" of http://compass-style.org/help/tutorials/contributing/#running-tests (ie running cucumber) still applies? I don't get cucumber on my system when running `bundle install`

PR #1474, opened 2013-12-09
table border: zero width's should not have units assigned

mprins:
I failed to find a test case for this so added a test + expectation


chriseppstein:
I'm starting to think that Sass should just omit the unit when outputting a number in compressed mode.

Oh look: https://github.com/nex3/sass/issues/649

mprins:
that would be nice too

PR #1477, opened 2013-12-11
save_blank_files added to config options

sparkida:
When using files in Sass as classes or models for mixins and variables, sometimes no content is rendered and an empty file will be saved. This is just a minor annoyance so I thought I'd provide a quick hook.


Anahkiasen:
Are you prefixing your files with `_` so that they're considered partials and not compiled ?

sparkida:
yes...now I am...ty :)

PR #1481, opened 2013-12-11
Fixes #1276: make :warning and :error messages log to stderr

### LINKED ISSUE #1276: Warnings emitted to stdout; corrupts stylesheet
When a warning is issued, it's output to stdout (in Compass::Logger.log) instead of stderr.

The warning is silently included in the stylesheet, when using tools like Brunch (via sass-brunch), causing a corrupt stylesheet.

Brunch is using sass via command line and quiet mode.  Quiet mode doesn't appear to be reflected in the logger at all either.

I'm a fan and user of Compass 0.12.2 (Alnilam).

jpap

ryanartecona:
Fixes #1276

### DIFF
--- a/cli/lib/compass/logger.rb
+++ b/cli/lib/compass/logger.rb
@@ -87,7 +87,8 @@ module Compass
     end
 
     def log_error(msg)
-      $stderr.write(msg << "\n")
+      $stderr.puts(msg)
+      $stderr.flush
     end
 
     # add padding to the left of an action that was performed.
--- a/cli/lib/compass/logger.rb
+++ b/cli/lib/compass/logger.rb
@@ -4,7 +4,8 @@ module Compass
 
     DEFAULT_ACTIONS = [:directory, :exists, :remove, :create, :overwrite, :compile, :error, :identical, :warning]
 
-    ERROR_ACTIONS = [:error, :warning]
+    LOG_METHODS = Hash.new {|h,k| :log}
+    LOG_METHODS.update(:error => :log_error, :warning => :log_error)
 
     COLORS = { :clear => 0, :red => 31, :green => 32, :yellow => 33 }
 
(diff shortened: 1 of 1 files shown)


chriseppstein:
I'm not sure this is going to fix #1276. There's no way to reproduce the original issue.

ryanartecona:
Right. I haven't reproduced the original issue, but I wanted to take a crack at an easy looking blocker. This at least sends errors/warnings to stderr, but coloring escape sequences still make it to both output streams. More would have to change to ensure color escapes only get sent to the streams being used. If we care about that, I can make that change here too.

This is all mainly because the original reporter seemed confident this was the issue, and I couldn't think of any other reason something like what he described would happen.

chriseppstein:
If the colors are a problem, they can pass `--boring` to disable them. I'm still unconvinced that they are using the compass command line tool at all for this. We never emit a file to stdout. I think they must be using the Sass command line tool which does operate on stdin/stdout. In which case, I think some of the warnings we were emitting were not going to stderr. but that this was not done through the logger.

ryanartecona:
Looks like you're right. From sass-brunch/src/index.coffee:

```
cmd = [
  'sass'
  '--stdin'
  '--load-path', @brunchConfig.paths.root
  '--load-path', sysPath.dirname path
  '--no-cache'
]
cmd.unshift 'bundle', 'exec' if @bundler

if @conf?.debug in ['comments', 'debug'] and not @brunchConfig.optimize
  hasComments = @conf?.debug is 'comments'
  cmd.push if hasComments then '--line-comments' else '--debug-info'

cmd.push '--scss' if /\.scss$/.test path
cmd.push '--compass' if @compass
```

So if we're sure this bug didn't touch the Compass logger, is this PR of any value, or just drop it?

chriseppstein:
I think we should just drop it. Sorry for not doing the due-diligence earlier. Sigh.

ryanartecona:
No worries. I was mostly trying to familiarize myself with the codebase, and thought it might help to knock out a simple issue. All was not for naught!

PR #1485, opened 2013-12-16
Snugug clearfix

### LINKED ISSUE #974: Default clearfix implementation has undesirable side effects
I’m tired of seeing focus styles, etc., get unceremoniously clipped like this (see left and bottom edges):

clipping

I’m going through some code and replacing `@include clearfix` with `@include pie-clearfix` which seems to have fewer side-effects.

Admittedly it is less concise, but it also doesn’t do, you know, _that_.

If I’m not missing anything, I’d like to suggest aliasing the `clearfix` mixin to `pie-clearfix`, and making the current `clearfix` available as `ppk-clearfix` or `oh-clearfix`.

Snugug:
Homygod Clearfix!

Update for #1151, will resolve #974 and #386.

The overflow clearfix method, while super small, in most cases isn't desired as it does more than just fix floating issues. A clearfix should, by default, only fix floating issues. If someone wants to use the overflow method, that should be called explicitly.

As for all of the extends, they're there to reduce overall output. I'm open to debating whether or not it should be enabled by default, but IMO most uses of clearfix don't happen from behind media queries and as such the extends should be fine.


Snugug:
I forgot tests! Sorry @chriseppstein! I'll do those this morning!

mirisuzanne:
Thanks @Snugug! This is a really cool feature, and I'd like to get it merged in. 

I think the failing test is not actually related to this change? The latest commit on master is passing, so try merging that in, and see if the tests pass. I'll leave my other thoughts inline.

core/stylesheets/compass/utilities/general/_clearfix.scss
```diff
@@ -2,43 +2,147 @@
...
-    visibility : hidden;
-  }
-  @include has-layout;
-}
-
-// This is an updated version of the PIE clearfix method that reduces the amount of CSS output.
-// If you need to support Firefox before 3.5 you need to use `legacy-pie-clearfix` instead.
-//
-// Adapted from: [A new micro clearfix hack](http://nicolasgallagher.com/micro-clearfix-hack/)
-@mixin pie-clearfix {
```
mirisuzanne, line 37: We'll need to deprecate the old `pie-clearfix` and `legacy-pie-clearfix` before we can remove them. That can happen with `@warn` and a re-direction to the most appropriate new mixin.

core/stylesheets/compass/utilities/general/_clearfix.scss
```diff
@@ -2,43 +2,147 @@
...
+// The type of clearfix to use by default
+// Default to false
+$default-clearfix-type: false !default;
+
+//////////////////////////////
+// Clearfix Mixins
+//////////////////////////////
+// [Overflow Clearfix](http://www.quirksmode.org/blog/archives/2005/03/clearing_floats.html)
+@mixin clearfix-overflow($extend: $default-clearfix-extend) {
+  @if not $extend {
```
mirisuzanne, line 21: While this use of `@extend` is an interesting feature to consider across compass more broadly, it requires a bit more thought. I'd prefer to discuss it as a separate feature, with an issue/pull-request of its own. As far as I'm aware, Compass currently has no output (including placeholder classes) unless explicitly requested — and I'd like to keep it that way. It's not hard for a user to create their own clearfix classes/placeholders as they want, and if we plan to provide that feature, it should be opt-in.

chriseppstein:

Snugug:
I'll do that.

chriseppstein:
@Snugug Thanks. I'm with eric on this one, can we leave out the extend parts for right now so that we can think about that as a more comprehensive strategy?

Snugug:
I’m cool with that too. There are a couple of different things I’d like to discuss in regards to a more comprehensive strategy, so we can have that discussion elsewhere.

chriseppstein:
#1746 appears to be an updated version. Thoughts?

PR #1500, opened 2014-01-03
Fix broken license link

danielbayerlein:
The link to the license was broken.


chriseppstein:
This is fixed in a different patch.

Issue #1501, opened 2014-01-03
Broken Link


joelworsham:
On Page: http://compass-style.org/

I was looking for:
http://green.cals.cornell.edu/

Cornell University

But instead I found:
404

PR #1507, opened 2014-01-08
Fix broken link in index page

mendozagioo:
Issue #1501


chriseppstein:
It's not clear that page is using compass now. I've just removed that link and replaced with with a different site in b45d23288db08a69ac4f35072c49344109fe8944.

PR #1520, opened 2014-01-19
IE7 requires html { font-size: 100%; } in establish-baseline()

JohnAlbin:
I was checking out the updated vertical-rhythm module and saw that `* html { font-size: 100%; }` is only added for IE 6. I seemed to remember that IE 7 also required that. So I looked it up…

IE 6 and 7 cannot resize fonts set with px. http://alistapart.com/article/howtosizetextincss#section3

IE 6 and 7 resize fonts weirdly when set with em. http://alistapart.com/article/howtosizetextincss#section4

Since IE7 doesn't understand the `* html` selector hack, we have to switch to the `*property` hack that both IE 6 and IE 7 know (but IE 8 and above do not.)

### DIFF
--- a/core/stylesheets/compass/typography/_vertical_rhythm.scss
+++ b/core/stylesheets/compass/typography/_vertical_rhythm.scss
@@ -96,15 +96,15 @@ $relative-font-sizing: if($rhythm-unit == px, false, true);
 // Establishes a font baseline for the given font-size.
 @mixin establish-baseline($font-size: $base-font-size) {
   $relative-size: 100% * ($font-size / $browser-default-font-size);
-
-  @if support-legacy-browser(ie, "6") and (not $relative-font-sizing) {
-    // IE 6 refuses to resize fonts set in pixels and it weirdly resizes fonts
-    // whose root is set in ems. So we set the root font size in percentages of
-    // the default font size, even if we are using absolute sizes elsewhere.
-    * html { font-size: $relative-size; }
-  }
   html {
     font-size: if($relative-font-sizing, $relative-size, $font-size);
+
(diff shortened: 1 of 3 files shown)


JohnAlbin:
Here's a more recent reference that shows it's both IE 6 and IE 7 that need the percent font.
https://github.com/necolas/normalize.css/blob/v1/normalize.css#L62

core/stylesheets/compass/typography/_vertical_rhythm.scss
```diff
@@ -95,16 +95,15 @@ $relative-font-sizing: if($rhythm-unit == px, false, true);
 
 // Establishes a font baseline for the given font-size.
 @mixin establish-baseline($font-size: $base-font-size) {
-  $relative-size: 100% * ($font-size / $browser-default-font-size);
```
mirisuzanne, line 98: Why was this line removed? `$relative-size` is still called on line 99, and now throws an error in the tests. This line should stay and be added to the new line 105 as well. No reason to duplicate that math.

mirisuzanne:
Thanks @JohnAlbin! Can you fix the broken variable, and get the tests passing? There are instructions for [writing and running the tests](https://github.com/chriseppstein/compass/blob/master/CONTRIBUTING.md#running-tests) in CONTRIBUTING.md file.

JohnAlbin:
Ok. The tests are "fixed". Except that they are still being reported as broken because the new tests depend on #1524, so that we can actually test when IE7 support is turned off.

chriseppstein:
#1524 was rejected, so can we get this rebased off of master?

chriseppstein:
I mean stable.

Issue #249, opened 2011-01-14
disable sprite folder prefix in selectors


chriseppstein:
Add an option to be able to take off the sprite folder prefix in sprite selectors.

scottdavis:
can you give an example im not 100% sure what you mean

scottdavis:
get me a use/test case and ill get this into 0.13

scottdavis:
i think this is possible by setting `$<%= name %>-sprite-base-class`

chriseppstein:
I don't know what I meant either.

rds:
This was actually a good idea – currently `@include all-icons-sprites` prefixes all the classes with `icons-`, it would be nice to have the option to remove the prefix completely so you can do `class="add icon"` instead of `class="icons-add icon"`.

friedhelm-langhof:
+1

rds:
I've modified the sprite importer template to utilise `$sprite-name-prefix` with an example. https://gist.github.com/rds/3766814

scottdavis:
feel like turning it into a pull request?

hagenburger:
+1

materliu:
+1

amelnikova-cn:
+1

PR #1521, opened 2014-01-19
Added support for the $<sprite-map>-prefix attribute.

mrinck:
With the $<sprite-map>-prefix attribute the default prefix (= sprite folder name) for the generated sprite classes can be overwritten.

My use case: In some cases we use our company branded icons as single images, in other cases they are merged in sprite sheets. It's easier for everyone if file names and sprite classes are identical. So i don't want the sprite classes to be prefixed with their including folder's name. In this case I set the  $<sprite-map>-prefix to null.
(This could also be the use case for Issue https://github.com/chriseppstein/compass/issues/249)

I found the existing $<sprite-map>-prefix attribute in the original code unused. So I hijacked it for my purposes.


chriseppstein:
This is a good patch, but we need to have it rebased on top of master after #1771 lands.

Issue #1497, opened 2013-12-30
EACCES Error on Compass watch


rpkoller:
I am running osx 10.8.5. with gem version 1.3.6 and ruby 1.8.7 (2012-02-08 patchlevel 358) [universal-darwin12.0]. Today i've removed all local gems. Afterwards i wanted to give thh Sass and Compass prerelease versions a spin alongside a certain gem. I've installed: 

```
sudo gem install compass --pre
sudo gem install modular-scale --pre
```

The local gem list looks like that: 

```
*** LOCAL GEMS ***

chunky_png (1.2.9)
compass (1.0.0.alpha.17)
compass-core (1.0.0.alpha.16)
compass-import-once (1.0.2)
ffi (1.9.3)
fssm (0.2.10)
json (1.8.1)
listen (1.1.6)
modular-scale (2.0.0.alpha3)
multi_json (1.8.2)
rb-fsevent (0.9.3)
rb-inotify (0.9.3)
rb-kqueue (0.2.0)
sass (3.3.0.rc.2)
```

But when i tried "compass watch" on a sample sass snippet i've received an error like in an previous alphas descibred in the issue here: https://github.com/chriseppstein/compass/issues/1406  . It's just odd that it happens on a "clean install" here. i've performed a compass watch --trace afterwards:

```
[danton@localhost][testung 2] compass watch --trace
>>> Compass is watching for changes. Press Ctrl-C to Stop.
/Users/danton/Desktop/Testung 2/scss/style.scss:6 DEBUG: 67.5px
identical css/style.css 
Errno::EACCES on line 1039 of /System/Library/Frameworks/Ruby.framework/Versions/1.8/usr/lib/ruby/1.8/fileutils.rb: Permission denied - /Library/Ruby/Gems/1.8/gems/compass-core-1.0.0.alpha.16/stylesheets/.listen_test
  /System/Library/Frameworks/Ruby.framework/Versions/1.8/usr/lib/ruby/1.8/fileutils.rb:1039:in `open'
  /System/Library/Frameworks/Ruby.framework/Versions/1.8/usr/lib/ruby/1.8/fileutils.rb:1039:in `touch'
  /System/Library/Frameworks/Ruby.framework/Versions/1.8/usr/lib/ruby/1.8/fileutils.rb:1033:in `each'
  /System/Library/Frameworks/Ruby.framework/Versions/1.8/usr/lib/ruby/1.8/fileutils.rb:1033:in `touch'
[...]

PR #1522, opened 2014-01-20
Fix of listen_test access issue when calling compass watch

impatient:
Related to ticket #1497

The way the listener was started, it included the compass gem stylesheet directory, which is read-only for default installs, thus getting the EACCES error. 

Couldn't find any existing unit tests that involved registering a framework.  I did run existing unit tests and verified they still worked.  I also verified that I'm no longer getting the permission issue.

I don't really do ruby.  Please review.

### DIFF
--- a/cli/lib/compass/configuration/adapters.rb
+++ b/cli/lib/compass/configuration/adapters.rb
@@ -70,19 +70,12 @@ module Compass
         Compass::Frameworks::ALL.each do |f|
           load_paths << f.stylesheets_directory if File.directory?(f.stylesheets_directory)
         end
-        importer = sass_options[:filesystem_importer] if sass_options && sass_options[:filesystem_importer]
-        importer ||= Sass::Importers::Filesystem
-        load_paths += resolve_additional_import_paths
-        load_paths.map! do |p|
-          next p if p.respond_to?(:find_relative)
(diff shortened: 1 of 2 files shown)


impatient:
See ticket #1518 to get a better idea of when this happens.

cli/lib/compass/configuration/adapters.rb
```diff
@@ -80,6 +80,18 @@ def sass_load_paths
...
+      def watch_load_paths
+        load_paths = []
+        importer = sass_options[:filesystem_importer] if sass_options && sass_options[:filesystem_importer]
+        importer ||= Sass::Importers::Filesystem
+        load_paths += resolve_additional_import_paths
+        load_paths.map! do |p|
+          next p if p.respond_to?(:find_relative)
+          importer.new(p.to_s)
+        end
+        load_paths
```
scottdavis, line 86: I would clean this up its unnecessary code duplication from the method above.

impatient:
Added suggested fixes and included Chris' info from the ticket.  I tried accessing the remove_redundant_directories method statically, but was unable.   Again.. not a ruby guy, can make other tweaks if needed.

chriseppstein:
The watcher was completely rewritten, this bug was fixed in sass.

PR #1523, opened 2014-01-20
Allow comma-delimited and nested lists for rem()

mirisuzanne:
In the past, `rem()` only accepted space-separated lists, one level deep. `rem(background-size, (25rem, 100% 32px))` would break. Now it doesn't.


mirisuzanne:
Whatever is failing here, it isn't my tests. :/

PR #1525, opened 2014-01-23
Webkit supports box-sizing

VinceCgto:


Anahkiasen:
That's not how prefixes are handled anymore so that shouldn't be necessary, check the master branch.

VinceCgto:
Ah yes, I see.

Still needs to be changed somewhere, though I'm not familiar with the current handling of prefixes.

Anahkiasen:
Latest Compass uses the Can I Use support data, but keep in mind the `-webkit-` prefix might not come from Chrome or Safari but might be added [for a mobile device](http://caniuse.com/#search=box-sizing).

chriseppstein:
This is data-driven in the new release.

PR #1527, opened 2014-01-24
Allow mixins outputting selectors to be used at the stylesheet root

JohnAlbin:
I was converting a CSS ruleset of:

```
::selection {
  background: #ca322c;
  color: white;
}
```

to use Compass, when I saw that the Compass mixin required a `* { }` wrapper. With Sass 3.3's `#{&}` capability, it should be possible to use any mixin that outputs a selector at the root level of the stylesheet.

Currently, doing this at the root of a stylesheet produces an error, but it shouldn't:

```
@include selection(#ca322c, white);
```

### DIFF
--- a/core/stylesheets/compass/css3/_selection.scss
+++ b/core/stylesheets/compass/css3/_selection.scss
@@ -11,11 +11,9 @@ $selection-support-threshold: $graceful-usage-threshold !default;
 // during selection: color and background-color. Firefox supports the
 // text-shadow property.
 //
-// At the stylesheet root, include the mixin within the * selector.
+// At the stylesheet root, include the mixin.
 //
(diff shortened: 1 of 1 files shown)


chriseppstein:
Script parent didn't get into Sass 3.3. Instead, I've added a custom function `at-stylesheet-root()` which returns `true` when `&` used to return `null`. Can you rework this patch to use that instead?

See: 5fb0b88144bcd2d139621c84da3de1cd7ac530a4

chriseppstein:
@JohnAlbin Also, this needs tests :)

chriseppstein:
@JohnAlbin Sass 3.4 is being released today. want to dust this off for 1.1?

PR #1528, opened 2014-01-24
Minor README corrections

lolmaus:


import-once/README.md
```diff
@@ -11,9 +11,9 @@ without compass in any Sass-based project.
 
 ## Installation
 
-Or add this line to your application's Gemfile if you have one:
+Either add this line to your application's Gemfile if you have one:
```
scottdavis, line 14: I am not sure `Either` makes since here. the `Or` definitely didn't belong. Should just be `Add this line....`

lolmaus, line 14: Did you read the whole thing?

Either add this line and this line, or install yourself with that. Perfect English.

Issue #1531, opened 2014-01-27
Wrong syntax in baseline-grid-background()


jonespen:
`baseline-grid-background()`, which uses the function `get-baseline-gradient()`, outputs `background-image: linear-gradient(bottom, #000000 5%, rgba(0, 0, 0, 0) 5%);` (among other vendor prefixes ofc). This is not according to the [linear-gradient specification](https://developer.mozilla.org/en-US/docs/Web/CSS/linear-gradient#Syntax), and the background is ignored by Chrome. The correct output should be `background-image: linear-gradient(to bottom, #000000 5%, rgba(0, 0, 0, 0) 5%);` (note _to bottom_)

mirisuzanne:
Thanks! Can you submit a pull request on the master branch?

chriseppstein:
PR was merged.

PR #1533, opened 2014-01-28
Fixed syntax for linear-gradient

jonespen:
... on master branch. Closes GH-1531


mirisuzanne:
Cool. The tests are failing because "bottom" (in the old syntax) should actually become "to top" in the new syntax. I don't know that direction is important for this gradient, but the mixin and tests need to match. You can either make the change here (probably the easiest), or update [the tests](https://github.com/chriseppstein/compass/blob/master/cli/test/fixtures/stylesheets/compass/css/grid_background.css) to get them passing.

jonespen:
Ah, my bad, it was actually off by 1px using `to bottom`. Changed to `to top`.

mirisuzanne:
Thanks!

Issue #102, opened 2010-03-16
Box Shadow: none


merbjedi:
I have a need to specify box-shadow: none in one of my styles. However, the `+box-shadow(none)` didnt like this. Is there a better way?

I had to just revert to using standard duplicated CSS3 styles:

```
box-shadow: none
-moz-box-shadow: none
-webkit-box-shadow: none
```

mirisuzanne:
I've been using +box-shadow("transparent") or +box-shadow(rgba(0,0,0,0)) to pull this off, but I agree that this could be useful. Or maybe +remove-box-shadow(). I'll take a look at patching that.

chriseppstein:
Allow box shadow to be turned off by @include box-shadow(none). Closed by 909fb92ad35f92cebf6240a7bb4b48294ce2e5aa.

DavidOliver:
Think this still needs to be added to the docs.

mirisuzanne:
Can you file a separate issue, or even better, submit a patch yourself? Thanks!

PR #1535, opened 2014-01-30
Added in a 'Possible Values: none' to the documentation for box-shadow, ...

RichardLitt:
...as per ticket #102 which was never closed.

### DIFF
--- a/doc-src/Gemfile.lock
+++ b/doc-src/Gemfile.lock
@@ -10,7 +10,7 @@ GIT
 PATH
   remote: ..
   specs:
-    compass (0.12.2.98929ea)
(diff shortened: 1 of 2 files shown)


RichardLitt:
I should say never resolved. The ticket itself was closed.

chriseppstein:
`none` is a legal value for the box shadow property. I feel like the existing documentation sufficiently describes the arguments accepted and adding this makes it feel like other values are not allowed.

chriseppstein:
But thank you for making the patch.

RichardLitt:
Sounds good. Thanks.

Issue #1306, opened 2013-05-31
Use sass's vendored version of Listen.


foiseworth:
Compass requires ~> 1 and Sass requires ~> 0.7. I'm not a ruby expert and I'm unsure if their is an easy method for overcoming this?

``` ruby
/Users/andrew/.rvm/rubies/ruby-2.0.0-p195/lib/ruby/site_ruby/2.0.0/rubygems/specification.rb:1193:in `block in activate_dependencies': can't satisfy 'listen (~> 1.0.0)', already activated 'listen-0.7.3' (Gem::LoadError)
    from /Users/andrew/.rvm/rubies/ruby-2.0.0-p195/lib/ruby/site_ruby/2.0.0/rubygems/specification.rb:1185:in `each'
    from /Users/andrew/.rvm/rubies/ruby-2.0.0-p195/lib/ruby/site_ruby/2.0.0/rubygems/specification.rb:1185:in `activate_dependencies'
    from /Users/andrew/.rvm/rubies/ruby-2.0.0-p195/lib/ruby/site_ruby/2.0.0/rubygems/specification.rb:1167:in `activate'
    from /Users/andrew/.rvm/rubies/ruby-2.0.0-p195/lib/ruby/site_ruby/2.0.0/rubygems/core_ext/kernel_gem.rb:48:in `gem'
    from /Users/andrew/.rvm/gems/ruby-2.0.0-p195/bin/compass:22:in `<main>'
    from /Users/andrew/.rvm/gems/ruby-2.0.0-p195/bin/ruby_noexec_wrapper:14:in `eval'
    from /Users/andrew/.rvm/gems/ruby-2.0.0-p195/bin/ruby_noexec_wrapper:14:in `<main>'
```

foiseworth:
Looks like @chriseppstein has already started the fix for this on the Sass repo:

https://github.com/nex3/sass/pull/777

foiseworth:
Compass will need to be upgraded to listen 1.1 to support Chris's fix

chriseppstein:
I'm on it. will be fixed today.

pshemek:
Hi, isn't it fixed yet? I'd need to use listen (1.2.3) in my gemset (Middleman app) but I can't...

scottdavis:
@chriseppstein

ghost:
I'm also having troubles with this. Compass 0.13.alpha.4 seems incompatible with guard-compass for similar reasons.

```
[...]

PR #1594, opened 2014-03-10
Be more optimistic re: compatible listen gem versions.

pprince:
Unless you know that Compass is incompatible with newer versions of ruby gem listen, please consider being less pessimistic about its version specifier.

Leads to conflicts bundling alongside, e.g., jekyll-1.4.3


pprince:
Looks like this is a duplicate of what #1306 has become.  Feel free to close it if so.

pprince:
Older, similar discussion in #1177 

Is the reason we depend on 1.1.0 that it supports Ruby 1.8?

If so, wouldn't that still work fine if you change it to 1.1 like in this PR?

pprince:
+1 vote for #1306; but, since it's already marked blocker, closing this.

PR #1597, opened 2014-03-11
Replace broken Cornell link with Causes.com

### LINKED ISSUE #1576: Link in "Brilliant people use Compass" leads to nowhere
On Page: http://compass-style.org/

I was looking for:
Site

But instead I found:
Site http://green.cals.cornell.edu/  is down, but link is available

lencioni:
On the Compass homepage, the link to Cornell University - CALS was
broken. I considered changing the link from
http://green.cals.cornell.edu to http://cals.cornell.edu, but that page
looked quite different than the thumbnail on the Compass homepage and I
was unable to verify that it actually used Compass. So instead, I
swapped it out for Causes.com, where we are making extensive use of
Compass.

Fixes #1576.


chriseppstein:
replaced with linkedin in b45d23288db08a69ac4f35072c49344109fe8944

PR #1598, opened 2014-03-12
Fixed a few typos.

KatieK2:
Fixed a few typos in spriting customization documentation.

PR #1601, opened 2014-03-13
Fixed a few typos.

KatieK2:
This is a dupe of #1598 (1 commit into chriseppstein:stable from KatieK2:patch-1) which I guess somehow got on to the wrong branch, which was having build problems. Hopefully this is the right place for it. Use this one - really for realz.


mirisuzanne:
Thanks!

PR #1602, opened 2014-03-14
calculate brightness() only for color values; fixes #1599

### LINKED ISSUE #1599: filter: brightness(1.1) is broken
Compass’ own [brightness() function](https://github.com/chriseppstein/compass/blob/master/core/stylesheets/compass/utilities/color/_brightness.scss) breaks the [official CSS3 filter brightness](http://www.w3.org/TR/2013/WD-filter-effects-1-20131126/#funcdef-brightness):
- Compass returns a percentage of a color: `brightness($color)` = 0% to 100%
- CSS3 expects a value between 0 and 2 or 0% and 200%: `brightness(1.1)` = make it 10% brighter 

I would prefer to rename Compass’ `brightness($color)` to `brightness-of($color)` and raise a warning if `brightness($value)` is called with a color value. What do you think?

---

If you have the same issue, here’s a work around:

``` scss
.whatever {
  @include filter(unquote("brightness(0.7)"));
}
```

hagenburger:
Fixed https://github.com/chriseppstein/compass/issues/1599.

### DIFF
--- a/core/stylesheets/compass/utilities/color/_brightness.scss
+++ b/core/stylesheets/compass/utilities/color/_brightness.scss
@@ -5,8 +5,16 @@
 // a color. Brightness is sometimes called luminance.
 //
 // Returns a number between 0% and 100%, where 100% is fully bright
-// (white) and 0% is fully dark (black).
+// (white) and 0% is fully dark (black) for color values.
+//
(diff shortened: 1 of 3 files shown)


chriseppstein:
merged to master.

PR #1613, opened 2014-03-20
Suppress unknown vendor prefixed properties for border-radius

mstoltenburg:
Remove vendor prefixes for -o-border-radius and -ms-border-radius. Those properties don't exist. This is already fixed when two values are passed but not for the default behaviour.


Anahkiasen:
Already fixed on master with the new prefixes system using usage data.

chriseppstein:
This is fixed in 1.0

PR #1631, opened 2014-03-29
Fixes wrong link to licence

camilonova:

PR #1633, opened 2014-04-04
Fix a broken link in README.markdown

koistya:


amitburst:
I think #1631 is the right way to go.

chriseppstein:
fixed elsewhere.

PR #1639, opened 2014-04-09
Added a note on how to install the preview version

deplorableword:
Hey,

Would be super if `gem install compass --pre` was listed on the compass-style.org somewhere. Eventually found it here http://compass-style.org/blog/2013/11/27/compass-versioning-change/ but figured it would be better on the actual install page :)


chriseppstein:
I merged just the doc change as 76ef05cbcafec7f63f2ce02ad73e924377aca28d.

Issue #1650, opened 2014-04-14
Assigning to global variable "$default" by default is deprecated


drywall:
Just updated various versions of compass/sass/susy, which all have vaguely different and conflicting version preferences. 

Running compass 1.0.0.alpha.19 against Sass 3.3.4, I get the following deprecation warnings: 

DEPRECATION WARNING on line 121 of /[path]/compass-core-1.0.0.alpha.19/stylesheets/compass/css3/_transition.scss:
Assigning to global variable "$default" by default is deprecated.
In future versions of Sass, this will create a new local variable.

DEPRECATION WARNING on line 28 of /[path]/compass-core-1.0.0.alpha.19/stylesheets/compass/css3/_text-shadow.scss:
Assigning to global variable "$default" by default is deprecated.

drywall:
Oh, and with the following as well:

If you want to assign to the global variable, use "$default: -compass-space-list(compact($default-text-shadow-h-offset $default-text-shadow-v-offset $default-text-shadow-blur $default-text-shadow-spread $default-text-shadow-color)) !global" instead.
Note that this will be incompatible with Sass 3.2.

chriseppstein:
This issue will go away in sass 3.4. Someone is creating a `$default` global and it's not compass.

PR #1651, opened 2014-04-14
Update _transition.scss

drywall:
Resolving sass deprecation notice (#1650)


chriseppstein:
This is because you must have a global variable named default.

PR #1652, opened 2014-04-14
Update _text-shadow.scss

drywall:
Resolve a deprecation notice from newer versions of sass, #1650.


chriseppstein:
This is because you have a global named default. Sass 3.4 is released today and this will be fixed automagically when the scoping rules change.

Issue #1550, opened 2014-02-07
1.0.0.alpha.17: transition with & without delay outputted incorrectly for webkit


thany:
Consider this sass:

```
@include transition(.2s ease-in opacity, .2s .2s ease-in visibility);
```

This will output css that is incorrect only for webkit:

```
-moz-transition: 0.2s ease-in opacity, 0.2s 0.2s ease-in visibility;
-o-transition: 0.2s ease-in opacity, 0.2s 0.2s ease-in visibility;
-webkit-transition: 0.2s ease-in opacity, 0.2s 0.2s ease-in;
-webkit-transition-delay: 0s, visibility;
transition: 0.2s ease-in opacity, 0.2s 0.2s ease-in visibility;
```

As you can see, other prefixes are outputted correctly.

Related testcase: http://sassmeister.com/gist/8861702

kroleg:
Hi guys. 
I have a fix for this https://github.com/kroleg/compass/commit/e7dbe9e0690ee96dd06eee16ca256efbaf16005c
But: 
1) It's incomplete (i don't how how to check if one of values passed to transition mixin is "cubic-bezier()")
2) Regression test doesn't pass. What is it for?
3) Contribution policy is a bit confusing for me.
Can you please give advice me, how to solve (1) and (2)? And what i need to do before a pull request?
Thanks!

eddie-ruva:
@kroleg were you able to fix your problems with (1) and (2)? I would like to take a look at how we can fix this but before starting to code I wanted to check with you if you were able to get something good working so we don't have to start from scratch. 

By taking a look at the code I think a quick solution for anyone that runs into the problem before a fix is submitted is to call the mixin in the following way:

``` scss
// transition: <property> <duration> <timing-function> <delay>
@include transition(opacity .2s ease-in, visibility .2s ease-in .2s );
```

eddie-ruva:
[Compass docs for css3 transitions](http://compass-style.org/reference/compass/css3/transition/) mention that the order of the parameters for creating a transition are: 

``` scss
@include single-transition($property, $duration, [$function], [$delay])
@include transition([$transition-1], [$transition-2], [$transition-3], [$transition-4], [$transition-5], [$transition-6], [$transition-7], [$transition-8], [$transition-9], [$transition-10])
```

Also the code in the transition mixin seems to be very couple to the order in which the parameters are passed:

``` scss
$property: nth($transition, 1);
[...]

PR #1678, opened 2014-05-16
Fixes output in transition mixing when giving parameters in any order

eddie-ruva:
In the issue #1550 it was reported that the transition mixing did not handle the parameters right when they were not provided in the following order:

``` scss
@include transition([<property> <duration> <timing-function> <delay>]);
```

In order to have the same behavior as the definition of the transition rule by the W3C we should be able to accept the parameters in any order but it is important to remember that _"the first value that can be parsed as a time is assigned to the transition-duration, and the second value that can be parsed as a time is assigned to transition-delay."_

Reference:
[W3C transitions definition](http://www.w3.org/TR/css3-transitions/#transition-shorthand-property)

### DIFF
--- a/core/stylesheets/compass/css3/_transition.scss
+++ b/core/stylesheets/compass/css3/_transition.scss
@@ -56,6 +56,27 @@ $transitionable-prefixed-values: transform, transform-origin !default;
   }
 }
 
+// Returns $transition-map which includes key and values that map to a transition declaration
+@function transition-map($transition) {
+  $transition-map: ();
+
+  @each $item in $transition {
+    @if is-time($item) {
+      @if map-has-key($transition-map, duration) {
+        $transition-map: map-merge($transition-map, (delay: $item));
+      } @else {
+        $transition-map: map-merge($transition-map, (duration: $item));
+      }
+    } @else if map-has-key($transition-map, property) {
+      $transition-map: map-merge($transition-map, (timing-function: $item));
+    } @else {
+      $transition-map: map-merge($transition-map, (property: $item));
+    }
+  }
+
+  @return $transition-map;
+}
+
 // One or more properties to transition
 //
 // * for multiple, use a comma-delimited list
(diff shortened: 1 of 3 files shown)


eddie-ruva:
I made the pull request with one test failing in _/core/test/integrations/projects/compass/sass/transition.scss_ I was not sure if this test was relevant because the expected output for it was wrong. 

``` scss

// Test
.regression-912 { @include transition((background-color 0.5s ease-in 0s, width 0.5s ease-out 0s, height 0.5s ease-in 0s, top 0.5s ease-out 0s)); }

// Expected output
.regression-912 {
  -moz-transition: background-color, 0.5s, ease-in, 0s width 0.5s ease-out 0s height 0.5s ease-in 0s top 0.5s ease-out 0s;
  -o-transition: background-color, 0.5s, ease-in, 0s width 0.5s ease-out 0s height 0.5s ease-in 0s top 0.5s ease-out 0s;
  -webkit-transition: background-color, 0.5s, ease-in, 0s width 0.5s ease-out 0s height 0.5s ease-in 0s;
  -webkit-transition-delay: top 0.5s ease-out 0s;
  transition: background-color 0.5s ease-in 0s width 0.5s ease-out 0s height 0.5s ease-in 0s top 0.5s ease-out 0s; }
```

Also the transition mixing does not expect a list as a parameter and this also made me think that probably this test is not right for this mixing. Should this test be passing?

chriseppstein:
Ugh. That was a regression in the code. I fixed it here: 3f9d9c1121ec842876bf58394f1472db1b366487

core/stylesheets/compass/css3/_transition.scss
```diff
@@ -56,6 +62,27 @@ $transitionable-prefixed-values: transform, transform-origin !default;
   }
 }
 
+// Returns $transition-map which includes key and values that map to a transition declaration
+@function transition-map($transition) {
+  $transition-map: ();
+
+  @for $i from 1 through length($transition) {
```
chriseppstein:
``` scss
@each $item in $transition {
   // ...
}
```

core/stylesheets/compass/css3/_transition.scss
```diff
@@ -56,6 +62,27 @@ $transitionable-prefixed-values: transform, transform-origin !default;
...
+    $item: nth($transition,$i);
+    @if is-time($item) {
+      @if map-get($transition-map, duration) {
```
chriseppstein:
``` scss
@if map-has-key($transition-map, duration) {
  // ...
}
```

core/stylesheets/compass/css3/_transition.scss
```diff
@@ -29,6 +29,7 @@ $default-transition-delay: null !default;
 
 $transitionable-prefixed-values: transform, transform-origin !default;
 
+$timing-functions: linear,ease,ease-in,ease-in-out,ease-out,step-start,step-end,steps,cubic-bezier;
```
chriseppstein: This sort of white-list always scares me because it's not future proof.

Also note that `steps` and `cubic-bezier` take arguments so checking for them in a list is not going to work. We need to iterate over the list of possible timing functions and do a `str-index` on the argument.

Personally, I'd rather not white list this at all. Like times, I think the first string should be considered the property and the second string considered a timing function.

chriseppstein:
Looks good. Please rebase against master to pick up the fix I made.

When these few issues are sorted out it will be good to merge.

chriseppstein:
Merged to master. Thanks!

PR #1679, opened 2014-05-16
Update install.haml

a-p-o:
``` bash
$ compass create myproject
```

Outputs

```
directory myproject/
directory myproject/sass/
directory myproject/stylesheets/
   create myproject/config.rb
   create myproject/sass/screen.scss
   create myproject/sass/print.scss
   create myproject/sass/ie.scss
   create myproject/stylesheets/print.css
   create myproject/stylesheets/screen.css
   create myproject/stylesheets/ie.css
```

So, the install guide should not read that `css` will be the default CSS output directory when in fact `stylesheets` is the default CSS output directory.

### DIFF
--- a/doc-src/content/install.haml
+++ b/doc-src/content/install.haml
@@ -53,8 +53,8 @@ body_id: install
   using
   %input#sassdir(placeholder="sass")
   for the sass source directory,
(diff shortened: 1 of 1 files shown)


chriseppstein:
Thanks. merged to master too.

PR #1691, opened 2014-05-29
Adds support for SVG generation in gradients with degrees

eddie-ruva:


chriseppstein:
It looks like this was explicitly disabled because the generated svg is bad.

``` svg
<?xml version="1.0" encoding="utf-8"?> <svg version="1.1" xmlns="http://www.w3.org/2000/svg"><defs><linearGradient id="grad" gradientUnits="userSpaceOnUse" x1="-45deg" y1="" x2="-45deg" y2="" gradientTransform = "rotate(-45)"><stop offset="0%" stop-color="#dddddd"/><stop offset="100%" stop-color="#aaaaaa"/></linearGradient></defs><rect x="0" y="0" width="100%" height="100%" fill="url(#grad)" /></svg>
```

`x1` and `x2` shouldn't be degrees those are supposed to be points in pixels I think.

So we need to look at how svg can represent a degree based svg file that looks like what the css would look like.

eddie-ruva:
@chriseppstein I've made a fix adding the gradientTransform="rotate(X)" but still the svg gradient produced is not the right one. It seems like we would need to also generate the x1,y1,x2,y2 for the degree gradients but I have not found a way to calculate this positions for these kinds of gradients. I've setup an example of what I am talking about in here http://codepen.io/eddie-ruva/pen/vogkc?editors=110 based on the css generated by 

``` scss

$svg-gradient-shim-threshold: 1%;
@import "compass/css3/images";

div {
  @include background(linear-gradient(35deg, red, green));
}
```

Do you have any idea on how we could calculate the x1,y1,x2,y2 for the degree gradients?

chriseppstein:
I ended up implementing this. The commit is 1f20e44f4d36e37ffc4244615a138b876f1368bb.

PR #1694, opened 2014-05-30
font-url helper with cdn-hosts and fingerprinting

sluceno:
When your project hosts it's own font files right now they are served from the app server. This PR allows the possibility to add them to the CDN network with a fingerprint, the same as the image-url helper does.


sluceno:
One of the failing tests it's because the font "font1.woff" does not exist. 
That's true the font does not exist in the test/fixtures/fonts. 

Previously was not failing because the test did not have to deal with the font file at all, now for the cache_buster needs to perform a File.open, and it fails.

What I would do here is to re-add the font files, from my opinion the file should exists.

sluceno:
Reading the documentation about asset_cache_buster [link](http://compass-style.org/help/tutorials/configuration-reference/) says that the second attribute of the block is the real_path of the file, which may or may not exists.

But in the [compute_cache_buster](https://github.com/sluceno/compass/blob/font-url_with_cdn_hosts_and_fingerprint/lib/compass/sass_extensions/functions/urls.rb#L276) it's expecting the file to exists doing a File.new, and passing a file object instead of a string as the documentation says.

Don't know which one is the correct one. Thanks!

chriseppstein:
We need to get this rebased on stable.

PR #1698, opened 2014-06-08
Update compass.gemspec

antonioparisi:
Update sass version on the gemspec


antonioparisi:
Sorry guys, I will test properly before to release it

Issue #1699, opened 2014-06-10
$legacy-support-for-mozilla: true isn't !default


shaneafsar:
I noticed that setting **$legacy-support-for-mozilla** to false on our project was not preventing legacy mozilla code from rendering (see **inline-block** mixin). Looking at the source, I then noticed that the !default flag isn't set, which means that setting the variable before compass is included in your SASS file doesn't work like it does with the **$legacy-support-for-ie** flags.

chriseppstein:
This variable isn't used in `1.0`.

PR #1700, opened 2014-06-10
$legacy-support-for-mozilla, IE legacy inline-block

### LINKED ISSUE #1114: Vertical align in inline-block mixin
Why the property $alignmen from inline-block mixin is not assigned to vertical-align in $legacy-support-for-ie block?

```
@mixin inline-block($alignment: $inline-block-alignment) {
  @if $legacy-support-for-mozilla {
    display: -moz-inline-stack; }
  display: inline-block;
  @if $alignment and $alignment != none {
    vertical-align: $alignment; }
  @if $legacy-support-for-ie {
    *vertical-align: auto;
    zoom: 1;
    *display: inline; } 
}
```

shaneafsar:
*Adds !default flag to $legacy-support-for-mozilla.
*Scopes inline-block legacy output for only IE6 & IE7, and removes the <IE7 vertical-align:auto default hack.

Closes GH-1114 and GH-1699.


chriseppstein:
This doesn't apply to 1.0.

PR #1703, opened 2014-06-17
Firefox *need* a unit. As none specified, defaulting to px

joe-mojo:
FF supports perspective rule unprefixed, but needs a unit. The one giving the closest result to the same unitless value in Chrome seems to be px. I think a mixin accepting unit as argument is needed.


chriseppstein:
This patch won't work on our new system. I've filed bug #1772 to track this.

Issue #1472, opened 2013-12-08
font-url and image-url ignore only_path


glebm:
`font-url` and `image-url` helps ignore `$only-path` argument and always wrap the url in `url(..)`, as per

https://github.com/chriseppstein/compass/blob/stable/lib/compass/sass_extensions/functions/urls.rb#L52-L55

See also: https://github.com/thomas-mcdonald/bootstrap-sass/issues/480

PR #1724, opened 2014-07-16
Fix asset-url("/path", true)

glebm:
Fix URL helpers bug #1472 and clean up duplication a bit.


chriseppstein:
Can we get this rebased on top of stable and submitted against that branch?

glebm:
Re-submitted as https://github.com/Compass/compass/pull/1774

Issue #1665, opened 2014-05-05
sprite_file function pathing bug


brewster1134:
I was having trouble using the sprite-file function with alpha19 and found a couple possible bugs...

Firstly, the function attempts to return a relative path from the generated images folder. It attempt to do this without ensuring that the two paths are both absolute.  If the `generated_images_path` is a relative path, or more likely, the `project_path` is set to the current directory (`.`), Pathname's prefix exception is thrown...

Secondly, why is the sprite file path bound to the generated images path anyways?  This would seemingly create relationship restrictions between the your project source, images sources, and generated images location.

chriseppstein:
Use `generated_images_dir` if it's a relative directory.

chriseppstein:
project_path gets expanded internally to an absolute path.

chriseppstein:
I don't understand the question here:

> Secondly, why is the sprite file path bound to the generated images path anyways? This would seemingly create relationship restrictions between the your project source, images sources, and generated images location.

can you give a better example of the issue?

brewster1134:
so first thing is `generated_images_path` gets used explicitly from `sprite_file`so it doesn't seem that `generated_images_dir` has any effect.

---

regarding my question about pathing... i discovered it seems to only become an issue when compass is run in the context of guard.  i have a test application setup [here](https://github.com/brewster1134/test/tree/compass_1665) on the `compass_1665` branch

in the Gemfile i have compass pointing to your current master.
there is also a commented out line pointing to my forked master.

the diff between the branches is [here](https://github.com/brewster1134/compass/compare/chriseppstein:master...master)

if you run guard with the gem pointing to `chriseppstein/compass`  you get a `TypeError: no implicit conversion of nil into String` coming from `Compass::Core::SassExtensions::Functions::ImageSize` trying to initialize a new `ImageProperties` with a `nil` file

if you switch the gem to point to `brewster1134/compass`, it works fine.  i hope this provides more info than my poorly worded question can provide :smile:

scottdavis:
@brewster1134 can you open a PR so i can review it more sainly?

brewster1134:
can you not see the diff from the compare view?

here is the PR as welll https://github.com/chriseppstein/compass/pull/1696

scottdavis:
Oh i could see the diff just fine the PR just helps me remember that i need to look into it and gives you credit if it is indeed the fix :)

Issue #1617, opened 2014-03-21
Syntax error: different prefix: "/" and "../media/img" for `sprite-file' 1.0.0.alpha19


jsonUK:
Hi,

I currently get the error:

```
Syntax error: different prefix: "/" and "../media/img" for `sprite-file'
        on line 107 of ../src/sass/_mixins.scss, in `use-sprite'
        from line 43 of ../src/sass/core/_icons.scss
        from line 68 of ../src/sass/style.scss
  Use --trace for backtrace.
```

Which is generated by this function:

```
/*
 * Retina Sprites for Compass
 * by:              Gaya Kessler
 * last update:     10/25/13
 *
 * Usage:
 * 1. create two folders in your image directory (in this case 'icons' and 'icons-2x').
 * 2. adjust the foldernames defined below if you use different names.
 * 3. create sprite images for pixel ratio 1 screens and put them in the first folder.
 * 4. create sprite images for pixel ratio 2 screens and put them in the second folder, use the same filenames.
 * 5. use the sprite-image in your Sass/Scss using: '@include use-sprite(<sprite-name>)'
 */

//first we'll define the folders where the sprites are and their layouts
$icons: sprite-map("png/*.png", $layout: smart);
$icons-2x: sprite-map("png2x/*.png", $layout: smart);

//Sprite mixin, works perfectly with standard defines
@mixin use-sprite($sprite) {
    background-image: sprite-url($icons);
    background-position: sprite-position($icons, $sprite);
    background-repeat: no-repeat;
    overflow: hidden;
    display: block;
    height: image-height(sprite-file($icons, $sprite));
    width: image-width(sprite-file($icons, $sprite));

    @media (-webkit-min-device-pixel-ratio: 2), (-o-min-device-pixel-ratio: 3/2), (min--moz-device-pixel-ratio: 2), (min-device-pixel-ratio: 2), (min-resolution: 144dpi) {
        background-image: sprite-url($icons-2x);
        background-size: (image-width(sprite-path($icons-2x)) / 2) (image-height(sprite-path($icons-2x)) / 2);
        background-position: round(nth(sprite-position($icons-2x, $sprite), 1) / 2) round(nth(sprite-position($icons-2x, $sprite), 2) / 2);
        height: image-height(sprite-file($icons-2x, $sprite)) / 2;
        width: image-width(sprite-file($icons-2x, $sprite)) / 2;
    }
}
```

Config file:

```
http_path = '/'
# css_dir = 'assets/stylesheets'
[...]

PR #1728, opened 2014-07-23
Fixes #1600 - Invalid path returned by sprite_file

### LINKED ISSUE #1600: Arguments to include all-[sprite]-sprites don't seem to work anymore
On `1.0.0.alpha.19` it seems I can't do this anymore :

``` scss
@include all-icon-sprites(true);
```

To generate the corresponding width and height of the sprites. Was this never intended to be in the public API ? I get this :

```
error public/app/sass/styles.scss (Line 7 of _base.scss: different prefix: "/" and "public/app/img/sprite/generated" for `sprite-file')
```

Removing the boolean fixes the problem. Works fine on `1.0.0.alpha.18`.

eddie-ruva:
When specifiying a generic_images_dir sprite_file would return the wrong
path to the original image because this was relative to the generic_
images_path instead of the images_path.

This should fix all problems related also to: #1617 and #1665

### DIFF
--- a/cli/lib/compass/sass_extensions/functions/sprites.rb
+++ b/cli/lib/compass/sass_extensions/functions/sprites.rb
@@ -120,9 +120,9 @@ module Compass::SassExtensions::Functions::Sprites
     verify_map(map, "sprite")
     verify_sprite(sprite)
     if image = map.image_for(sprite.value)
-      image_path = Pathname.new(File.expand_path(image.file, Compass.configuration.images_path))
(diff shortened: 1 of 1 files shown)


lifeiscontent:
@chriseppstein this would be really nice to get in..

chriseppstein:
@eddie-ruva Thanks for the patch!
@lifeiscontent nice is what I do.

lifeiscontent:
:+1:

PR #1746, opened 2014-08-05
Allow configuration of which method "clearfix" uses

alanhogan:
The designers I know and have worked with all know what a clearfix is, and I think many assume that a project will contain a `clearfix` mixin, whether it comes from Compass or not. 

I do not encourage use of the `overflow: hidden` clearfix as I dislike its side-effects.

That is to say, I see real value in changing the default clearfix implementation in our project, as opposed to asking everyone to remember to use a different clearfix mixin.

This should enable that.

I can/will contribute documentation as well if this PR is likely to be accepted…?


alanhogan:
A tale of two clearfixes

### Button group implemented with floats & `clearfix`

screen shot 2014-08-07 at 1 06 12 pm

screen shot 2014-08-07 at 1 06 32 pm

---

### Button group implemented with floats & `pie-clearfix`

screen shot 2014-08-07 at 1 05 49 pm
screen shot 2014-08-07 at 1 05 36 pm

chriseppstein:
There was a similar effort in #1485. @Snugug Is this basically what the patch should be now?

Snugug:
Similar idea, but different clearfixes. Mine rewrite most of the clearfixes in favor of more modern approaches for each level of support and choose what clearfix to use based on level of browser support needed.

alanhogan:
@Snugug Do you have any interest in extending this PR to accomplish your goals, as well? Maybe add  those additional implementations and one additional meta-clearfix that selects one based on support needs? If so, feel free to send a PR to [alanhogan/compass](https://github.com/alanhogan/compass) or  maybe just open a new PR here using my branch as a starting point…?

Snugug:
@alanhogan When I have some time I'll take a look.

alanhogan:
Closing as Compass is no longer in active development

PR #1751, opened 2014-08-08
Add support for original webkit gradient syntax

eddie-ruva:
- Adds support for the linear gradients. 
- Missing support for radial gradient due to poor backwards compatibility. 

cc/ @chriseppstein


chriseppstein:
Thanks, @eddie-ruva.

PR #1774, opened 2014-08-19
Fix asset-url("/path", true)

glebm:
Same as PR #1724, but rebased with respect to the latest stable.

### DIFF
--- a/core/lib/compass/core/sass_extensions/functions/urls.rb
+++ b/core/lib/compass/core/sass_extensions/functions/urls.rb
@@ -30,12 +30,7 @@ module Compass::Core::SassExtensions::Functions::Urls
         Compass.configuration.http_root_relative(Compass.configuration.css_dir)
       end
 
-      path = "#{http_stylesheets_path}/#{path.value}"
-      if only_path.to_bool
-        unquoted_string(clean_path(path))
-      else
-        clean_url(path)
-      end
+      sass_clean_url "#{http_stylesheets_path}/#{path.value}", only_path
     end
   end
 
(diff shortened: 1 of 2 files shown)


core/lib/compass/core/sass_extensions/functions/urls.rb
```diff
@@ -241,12 +228,8 @@ def generated_image_url(path, cache_buster = bool(false))
   # Emits a path, taking off any leading "./"
   def clean_path(url)
     url = url.to_s
-    url = url[0..1] == "./" ? url[2..-1] : url
-  end
-
-  # Emits a url, taking off any leading "./"
-  def clean_url(url)
-    unquoted_string("url('#{clean_path(url)}')")
+    url = url[2..-1] if url.start_with?('./')
```
glebm, line 231: Why? Single quoted strings indicate simple non-interpolated strings.

Arcovion, line 231: Yea, you should always prefer single quoted strings. Also, line length 80? Really?
The styleguide this bot uses [is here](https://github.com/thoughtbot/guides/tree/master/style) - it's all very disagreeable, no vertical alignment, acronyms without capital letters, avoid conditional modifiers,  avoid bang methods?! It's not rubyish.

Suggest switching to RuboCop which is a more customisable, relaxed analyser that won't litter with comments and can automatically fix code by itself.

chriseppstein, line 231: We're just testing out this tool. We use rubocop on Sass and I liked it. But getting the code to comply is a TON of work.

Anahkiasen, line 231: @chriseppstein What about [Scrutinizer](https://scrutinizer-ci.com/) ? It aggregates various CI tools for PHP, Ruby, JS, and Python. It's quite powerful. Supports pull requests statuses and stuff.

glebm:
@chriseppstein Does anything else need to be done on my side for this to be merged?

chriseppstein:
I just re-wrote all our url handling. Grab compass 1.1.0.alpha.0 and let me know if it's working correctly for you.

glebm:
I've opened a separate PR with just the test that was previously failing but now passes with the current Compass master. https://github.com/Compass/compass/pull/1812

PR #1780, opened 2014-08-21
Add the 'unicode-range' option to font-face

mprins:
This PR adds the `unicode-range` option to `font-face`. This allows for minimising font downloads by specifying which characters may be found in a a font file, though currenlty only supported in Blink based browsers (Chrome/Opera/...)

For the lowdown see Jake Archibald's blogpost: http://jakearchibald.com/2014/minimising-font-downloads/


mprins:
closing as "Compass is dead"

PR #1793, opened 2014-08-25
Asset collections

chriseppstein:
This adds the ability to add multiple asset collections to compass. For each collection there is basically a set of asset options like you would use by default. This effectively forms a path list of lookup locations for images and fonts and sass files. The sass files in the project form a "default" collection of assets which allows the asset resolution code to be homogenous. URL construction and resolution is greatly simplified and dried up compared to before.

There are a couple of small backwards incompatibilities introduced in this patch. I'm trying to decide if this is worth a major version bump or if I should go through a deprecation phase first:
1. `http_path = :relative` is removed in favor of the long time preferred option `relative_assets = true`
2. The url passed to `asset_host` and `asset_cache_buster` now includes the full root-relative path (instead of being relative to the asset's http directory). This allows for better disambiguation and let's the relative path operation work correctly even when asset_cache_buster completely changes the url of the asset.
3. Images that are missing are now always an error instead of only sometimes being an error depending on configuration.


chriseppstein:
This landed on master

tdreyno:
This is a really cool feature. With the nightmare that are Sprockets gems, this might be a cool thing for us to support in Middleman as well.

PR #1797, opened 2014-08-27
Remove .sass-cache after creating project

hungtatai:
When compass create a new project, it'll compile this project first. At that time, however, the cache_path is point to `#{working directory}/.sass-cache`, not `#{project_path}/.sass-cache`. So I simply remove cache folder after creating project.


chriseppstein:
Actually, the right fix here is to create the sass cache in the project directory.

hungtatai:
I agree with you, @chriseppstein . So this time, I update cache_path before updating project to fix it. :)

PR #1839, opened 2014-09-30
Add related values to user-select mixin

seidior:
These related values, **-webkit-touch-callout** and **-webkit-tap-highlight-color**, control related values to user-select that should be part of the expected behavior when using the mixin to disable selection of texts or objects.


chriseppstein:
This breaks tests and doesn't fix them.

Furthermore, if someone isn't supporting a webkit browser then this should be output at all.
Use the `with-prefix` mixin to target this output for webkit browsers.

Can you iterate on this pull request?

PR #1849, opened 2014-10-13
Removed deprecated warning: "Webkit only supports pixels for the start and end stops for radial gradients"

oori:
See note at:   https://developer.apple.com/library/safari/documentation/InternetWeb/Conceptual/SafariVisualEffectsProgGuide/Gradients/Gradient.html#//apple_ref/doc/uid/TP40008032-CH10-SW15

```
The -webkit-linear-gradient and webkit-radial-gradient properties require iOS 5.0 or later, or Safari 5.1 or later on the desktop. 
If you need to support earlier releases of iOS or Safari, see “Prior Syntax (-webkit-gradient).”
```

It's not really clear why this warning is there in first place, as even the "Prior Syntax" has an example with start/stop using percentage:   

```
-webkit-gradient(radial, 20% 20%, 0px, 20% 20%, 60px, from(white), to(black))
```


chriseppstein:
Percentages were allowed for the in-between stops but the start and end stops had to be specified in pixels. If you don't want to see this warning, you just tune your browser support so that original webkit syntax isn't output anymore.

PR #1852, opened 2014-10-15
Update _images.scss

Hooter:
mixin filter-gradient() does not work correctly in different versions of IE.
and need to IE9 filter: none;


chriseppstein:
This was explicitly removed a while ago: https://github.com/Compass/compass/pull/447

What problem are you seeing?

PR #1865, opened 2014-11-02
Missing indentation

philgruneich:
A code example was missing 2 spaces on its indentation and Markdown was parsing it as regular text.


chriseppstein:
Thanks!

PR #1884, opened 2014-11-25
Don't puke if error.sass_filename returns nil.  Fixes #1883

### LINKED ISSUE #1883: sass_compiler.rb should handle the case where the sass_filename method returns nil
In sass_compiler.rb's `when_compilation_error` method, the code checks to see if the error object responds to [sass_filename](https://github.com/sass/sass/blob/84e55254674508a4006517fff636f750da39ed64/lib/sass/error.rb#L65).  When this error is an instance of an SyntaxError this will always evaluate to true, but the docs for sass_filename make it clear that this method can return nil.

I'm not 100% sure why it's returning nil in my particular case, but it is.  Regardless of why it's nil, this should probably be handled by compass correctly.  As it is now, this nil value is passed to relativize [here](https://github.com/Compass/compass/blob/350bcaa544f594bca972aaa29a9bdfddceee5d4f/cli/lib/compass/sass_compiler.rb#L116) which pukes on the nil value, preventing any meaningful error about the actual compilation error from being displayed.

I eventually traced the error to be a compilation error due to a non-ascii character in a template that got included by a changed template (this happens during compass watch).  FWIW, in this particular case the sass_file variable passed to when_compilation_error is perfectly fine.

mikesnare:
This fixes the issue in #1883

### DIFF
--- a/cli/lib/compass/sass_compiler.rb
+++ b/cli/lib/compass/sass_compiler.rb
@@ -113,10 +113,10 @@ class Compass::SassCompiler
   def when_compilation_error(error, sass_file, css_file, sourcemap_file)
     @error_count += 1
     if error.respond_to?(:sass_filename)
-      if relativize(error.sass_filename) == relativize(sass_file)
+      if !error.sass_filename.nil? && relativize(error.sass_filename) == relativize(sass_file)
         logger.record :error, "#{relativize(error.sass_filename)} (Line #{error.sass_line}: #{error.message})"
       else
-        logger.record :error, "#{relativize(sass_file)} (Line #{error.sass_line} of #{relativize(error.sass_filename)}: #{error.message})"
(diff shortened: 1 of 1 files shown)


mikesnare:
Build failures have been going on for a while now and have nothing to do with this PR:

```
    And I shutdown the other process # features/step_definitions/command_line_steps.rb:89
      No such file or directory - /tmp/last_result.compass_test.txt (Errno::ENOENT)
      ./features/step_definitions/command_line_steps.rb:92:in `read'
      ./features/step_definitions/command_line_steps.rb:92:in `/^I shutdown the other process$/'
      features/command_line.feature:145:in `And I shutdown the other process'
```

and 

```
ProjectsTest#test_compass:
TypeError: NilClass can't be coerced into Fixnum
    kernel/common/numeric.rb:168:in `math_coerce'
    kernel/common/numeric.rb:183:in `redo_coerced'
    kernel/bootstrap/fixnum.rb:34:in `*'
```

mikesnare:
Fixed by @chriseppstein in #1883

PR #1892, opened 2014-12-05
Update the docs to use the sass --compass flag

### LINKED ISSUE #1800: Using Compass without command line tools appears not to work
Maybe this is my bad, but we're attempting to follow the instructions at [https://github.com/Compass/compass/blob/stable/compass-style.org/content/help/index.haml#L49](content/help/index.haml#L49) where it says to @import "compass/configuration" and run the `sass -r compass-core scss:css` command.

The problem is that sass cannot find "compass/configuration," despite including the compass-core library.

Either the docs or wrong, or something is broken.  Please advise.

Ruby version: 1.9.3p194
Compass version: 1.0.1 (Polaris)
Sass version: 3.4.2 (Selective Steve)

navarr:
Fixes #1800

### DIFF
--- a/compass-style.org/content/help/index.haml
+++ b/compass-style.org/content/help/index.haml
@@ -52,7 +52,7 @@ layout: site
         EOF
         # Add to the top of each sass file: @import "project-setup";
         # Compile using the Sass command line:
-        $ sass -r compass-core --update sass:css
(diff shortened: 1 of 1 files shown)


navarr:
Wrong branch.

PR #2003, opened 2015-07-31
Set stretch mixin to not set unspecified values. Closes #2002

### LINKED ISSUE #2002: Mixin Stretch sets value 0 on unspecified values
div { @include stretch($offset-top:5px); }

Output:

div { position: absolute;top: 5px;bottom: 0;left: 0;right: 0; }

Expected:

div { position: absolute;top: 5px; }

Setting the 0 values causes unexpected errors in layout because these values were intentionally to be undefined or inherited.

dcondrey:
Stretch mixin should not set 0 values for positions not specified.  If I use stretch and set top, right, left, but don't set bottom.  Mixin sets bottom to 0 whereas my intention was to leave bottom undefined.

Set parameters to be optional rather than assigned default values.

Closes #2002

### DIFF
--- a/core/stylesheets/compass/layout/_stretching.scss
+++ b/core/stylesheets/compass/layout/_stretching.scss
@@ -15,10 +15,10 @@
 
 // shorthand to stretch element height and width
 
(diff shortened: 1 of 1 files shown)


cimmanon:
This introduces backwards incompatible changes.

PR #2075, opened 2016-04-01
V2

mrmrs:
Put together a first pass at a direction for v2. Think it's ready for public review.


scottdavis:
IM pretty sure you didn't mean to open this PR since compass is not called `Basecss`

PR #2095, opened 2016-08-04
Changed the header of Compass

JamesVanWaza:


scottdavis:
We have already merged in a notice about support I think one that removes documentation links is a bit extreme.

JamesVanWaza:
@scottdavis the reason I did that is so that the issues can no longer be added, is there a way for you to close all issues?

scottdavis:
I don't think closing the issues gains anything. It's an active record of known issues in case someone does have an issue if they are running compass in a ongoing or legacy project.

JamesVanWaza:
Ok. @scottdavis Then make the Header Larger so that it is easily seen.

scottdavis:
i changed the site description header.

PR #2119, opened 2017-03-02
fix UMDF donation url

jakesylvestre:
Currently, the umdf donation url points to http://umdf.org/compass which does not exist anymore.  This pull request changes the url to http://umdf.org/ so would be donors arrive at the home page.


jakesylvestre:
Closing since as per the readme, this is no longer supported and the last commit was 4 years ago.

Issue #2135, opened 2018-01-23
compass-import-once gem not compatible with sass 3.5


FooBarWidget:
I'm trying to get Middleman 4 to work with Bootstrap 4 using [https://github.com/twbs/bootstrap-rubygem](bootstrap-rubygem). However:

 * bootstrap-rubygem depends on sass >= 3.5.2.
 * compass-import-once (which Middleman 4 depends on) depends on sass (< 3.5, >= 3.2).

Please update compass-import-once to be compatible with sass 3.5.

davidmyers9000:
Same here

davidmyers9000:
Same here
@scottdavis @mirisuzanne

Until fixed, the 4.0.0.beta version of the gem is a decent workaround
`gem 'bootstrap', '~> 4.0.0.beta'`

PR #2138, opened 2018-05-04
Support Sass 3.5

HamptonMakes:
I'm not sure why we are stopping using the "latest" version.

### DIFF
--- a/core/compass-core.gemspec
+++ b/core/compass-core.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_dependency "sass", ">= 3.3.0", "< 3.5"
+  spec.add_dependency "sass", ">= 3.3.0"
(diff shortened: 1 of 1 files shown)


jonathanhefner:
[`import-once/compass-import-once.gemspec`](https://github.com/Compass/compass/blob/7cb58273f96e46e3bf3a1413d555cc105d5f83b6/import-once/compass-import-once.gemspec#L21) also needs this change, as per #2135. 👍 

Also [`cli/compass.gemspec`](https://github.com/Compass/compass/blob/7cb58273f96e46e3bf3a1413d555cc105d5f83b6/cli/compass.gemspec#L16), I suppose, going by Compass/compass@7cb58273.

junaruga:
@hcatlin could you update below gemspec files to update not only `compass-core` but also `compass` and `compass-import-once`?

```
./cli/compass.gemspec
./core/compass-core.gemspec
./import-once/compass-import-once.gemspec
```

Thank you.

junaruga:
See this repo's top page.

> https://github.com/Compass/compass
> Compass is no longer actively maintained.

We should use the compass alternative.

PR #2167, opened 2023-04-24
In Rails 3.2 the method File#exists? is removed

dima4p:


solaris007 (changes requested):
You're missing many more instances of `File.exists?`: https://github.com/search?q=repo%3ACompass%2Fcompass+%22File.exists%3F%22&type=code

PR #2170, opened 2024-10-08
Replace File.exists? with File.exist?

### LINKED ISSUE #2169: NoMethodError on line ["89"] of /usr/local/bundle/gems/compass-1.0.3/lib/compass/configuration/helpers.rb: undefined method `exists?' for class File
When I create this Dockerfile:

```
FROM ruby:alpine

RUN apk add --no-cache build-base && \
  gem update --system && \
  gem install compass

VOLUME /src

WORKDIR /src

ENTRYPOINT [ "compass" ]
```

Then I run:

```
docker build -t my-compass .

docker run --rm -v "$(pwd)"/example01:/src my-compass compile /src/ --force --trace
```

If my "$(pwd)"/example01/scss/style.scss contains:

```
.facets-select-wrapper {
  position: relative;

  .facets-list {
    position: absolute;
    top: 100%;
    left: 0;
    display: inline-block;
    border: 1px solid #dedede;
    padding: 10px;
    max-height: 300px;
    overflow: scroll;
    z-index: 100;
    background: #fff;
    transition: 0.2s all;
    opacity: 0;
    visibility: hidden;
  }
  .facets-trigger {
    display: inline-block;
  }
  .facets-trigger:hover > .facets-list {
    opacity: 1;
    visibility: visible;
  }
}
```

I would expect this to generate css files, however i see the following:

```
NoMethodError on line ["89"] of /usr/local/bundle/gems/compass-1.0.3/lib/compass/configuration/helpers.rb: undefined method `exists?' for class File
```

jeffbyrnes:
Fixes #2169
Closes #2167

### DIFF
--- a/cli/features/step_definitions/command_line_steps.rb
+++ b/cli/features/step_definitions/command_line_steps.rb
@@ -114,11 +114,11 @@ Then /^a directory ([^ ]+) is (not )?created$/ do |directory, negated|
 end
  
 Then /an? \w+ file ([^ ]+) is (not )?removed/ do |filename, negated|
-  File.exists?(filename).should == !!negated
+  File.exist?(filename).should == !!negated
 end
 
 Then /an? \w+ file ([^ ]+) is (not )?created/ do |filename, negated|
-  File.exists?(filename).should == !negated
+  File.exist?(filename).should == !negated
 end
 
 Then "the following files are reported removed:" do |table|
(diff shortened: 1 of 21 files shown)


damienmckenna (approved):
Thanks for the fix!

sffc:
Please merge this and release this. I have some projects that use compass and they don't work on Ruby 3.2+ which is the default version in package managers now.

karol-haltenberger:
Dirty "fix" until this is merged.
`find /var/lib/gems/3.3.0/gems/ -type f -name *.rb -exec sed -i 's/File.exists/File.exist/g' {} +`

